Issue #3, opened 2022-02-08
TST: testing error with pytest


tylerjereddy:
I've finally managed to avoid various CUDA driver/runtime errors, but now when I run `pytest` I see all tests fail in a manner similar to this:

```
____________________________________________________________________________________________________________________________________________________ TestViews.test_v3d ____________________________________________________________________________________________________________________________________________________

self = <tests.test_views.TestViews testMethod=test_v3d>

    def test_v3d(self):
>       pk.parallel_for(self.range_policy, self.functor.v3d)

tests/test_views.py:186: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pykokkos/interface/parallel_dispatch.py:126: in parallel_for
    runtime_singleton.runtime.run_workunit(
pykokkos/core/runtime.py:85: in run_workunit
    return self.execute(workunit, module_setup, members, policy=policy, name=name, **kwargs)
pykokkos/core/runtime.py:119: in execute
    module = self.import_module(module_setup.name, module_setup.path)
pykokkos/core/runtime.py:148: in import_module
    module = importlib.util.module_from_spec(spec)
<frozen importlib._bootstrap>:556: in module_from_spec
    ???
<frozen importlib._bootstrap_external>:1166: in create_module
    ???
[...]

PR #4, opened 2022-02-13
WIP, TST: install tests with package

tylerjereddy:
* install the tests with `pykokkos` so that it is
possible to run i.e., `pytest --pyargs pykokkos`
for an installed version of `pykokkos` from any
path on the filesystem

* this doesn't work yet though, still WIP; I think
there are some issues with relative paths maybe,
and some of the other points I brought up in gh-3
may make this difficult for now.

* there are other testing scripts that would probably
need to be moved too, like `tests.sh` I suppose


tylerjereddy:
In meeting today we decided to delay this until `pytest` can be used from the command line (import tricks with pykokkos).

PR #7, opened 2022-02-16
MAINT: use a larger type for simple_reduce

### LINKED ISSUE #6: MAINT: overflow in 02_simple_reduce.py
For the `02_simple_reduce.py` example, I seem to get overflow if I do i.e., `SquareSum(90000)`. If it isn't too distracting to the example, perhaps switching to `pk.double` for the `self.total` type would be acceptable, and better match the type of the internal accumulator of the sum in any case.

tylerjereddy:
Fixes #6 (well up to a smaller value at least..)

* use a larger type for the total variable
in `02_simple_reduce.py` to prevent overflow
at larger square sum values, and to match
the type of the accumulator that is used

* this seems to allow `9000` to be handled,
though not `90000`, based on comparison
with i.e., `np.sum(np.arange(9000) ** 2)`

* I also noticed that this adjustment
only helps if I manually purge the `pk_cpp`
folder (no recompile on the variable
type change?)

### DIFF
--- a/examples/kokkos/02_simple_reduce.py
+++ b/examples/kokkos/02_simple_reduce.py
@@ -5,7 +5,7 @@ import pykokkos as pk
 class SquareSum:
     def __init__(self, n):
         self.N: int = n
-        self.total: int = 0
+        self.total: pk.double = 0
 
     @pk.main
     def run(self):


examples/kokkos/02_simple_reduce.py
```diff
@@ -5,7 +5,7 @@
 class SquareSum:
     def __init__(self, n):
         self.N: int = n
-        self.total: int = 0
+        self.total: pk.double = 0
```
tylerjereddy, line 8: Is there a larger integer type we could use instead?

NaderAlAwar: There's `pk.int64`, but it won't work here since for the style with `pk.main`, we pack all the reduction results into a single view and return that to Python. By default that view is of type `double`, so assigning an `int64_t` will result in a compilation error.

NaderAlAwar (approved):
This looks good to me

PR #10, opened 2022-03-22
TST: assert_allclose for test_float_sum

### LINKED ISSUE #9: TST: test_float_sum failure
This test fails for me locally and indeed you should almost never assert exact equality between floating point values, so I'll open a PR with a simple fix shortly:

```
____________________________________________________________________________________________________________________ TestKokkosFunctionsTranslator.test_float_sum _____________________________________________________________________________________________________________________

self = <tests.test_kokkosfunctions_translator.TestKokkosFunctionsTranslator testMethod=test_float_sum>

    def test_float_sum(self):
        expected_result: float = self.threads * (self.f_1 + self.f_2)
        result: float = pk.parallel_reduce(self.range_policy, self.functor.add_floats)
    
>       self.assertEqual(expected_result, result)
E       AssertionError: 340.0 != 340.0000000000003

tests/test_kokkosfunctions_translator.py:168: AssertionError

```

tylerjereddy:
Fixes #9

* use the NumPy testing function `assert_allclose()`
for the floating-point assertion in `test_float_sum()`
because requiring exact floating point equality was
causing falures for me on one of our `clx-volta` nodes

* I'm assuming that a dependency on NumPy for the
testsuite should be "ok" (test-time deps are usually
less serious than runtime deps, etc.)

### DIFF
--- a/tests/test_kokkosfunctions_translator.py
+++ b/tests/test_kokkosfunctions_translator.py
@@ -166,7 +166,7 @@ class TestKokkosFunctionsTranslator(unittest.TestCase):
         expected_result: float = self.threads * (self.f_1 + self.f_2)
         result: float = pk.parallel_reduce(self.range_policy, self.functor.add_floats)
 
-        assert_allclose(expected_result, result)
+        assert_allclose(result, expected_result)
 
     def test_nested_sum(self):
         expected_result: int = self.threads * (self.i_1 + self.i_2)
--- a/tests/test_kokkosfunctions_translator.py
+++ b/tests/test_kokkosfunctions_translator.py
@@ -1,6 +1,7 @@
 import unittest
 
 import pykokkos as pk
+from numpy.testing import assert_allclose
 
 
 # Tests translation of KOKKOS_FUNCTIONS to C++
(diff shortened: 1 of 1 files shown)


tylerjereddy:
This allows the `pytest` run to pass for me locally: `89 passed, 3 warnings in 22.81s` using the approach in https://github.com/kokkos/pykokkos/issues/3#issuecomment-1075606752

tests/test_kokkosfunctions_translator.py
```diff
@@ -165,7 +166,7 @@ def test_float_sum(self):
         expected_result: float = self.threads * (self.f_1 + self.f_2)
         result: float = pk.parallel_reduce(self.range_policy, self.functor.add_floats)
 
-        self.assertEqual(expected_result, result)
+        assert_allclose(expected_result, result)
```
tylerjereddy, line 169: note that for the NumPy testing functions we should also really use the order `actual, expected`, though I believe the current convention in pykokkos does largely the reverse, so I didn't change that yet..

NaderAlAwar: Good point. Should we open another PR? Or do you want to fix it here?

PR #12, opened 2022-03-23
MAINT: rm duplication in bindings module

tylerjereddy:
* the tests seem to pass without this duplication
detected by `mypy`; I haven't yet checked if the
current tests flush this code path though

* the original `mypy` complaint was:
  - `pykokkos/core/translators/bindings.py:179: error: Name "view_type" already defined on line 173`

Incidentally, I was kind of thinking a scoped `mypy` check might be an easy way to get CI rolling (i.e., start with a tiny bit of static type analysis, then ramp up to `runtests.py` and all the compiled dependencies).

### DIFF
--- a/pykokkos/core/translators/bindings.py
+++ b/pykokkos/core/translators/bindings.py
@@ -176,7 +176,6 @@ def generate_copy_back(members: PyKokkosMembers) -> str:
             continue
 
         # skip views with user-set memory spaces
-        view_type: cppast.ClassType = members.views[cppast.DeclRefExpr(v)]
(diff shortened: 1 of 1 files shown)


NaderAlAwar:
Yeah a `mypy` check would be great. Does it report anything else, like type errors? Do we want to fix all those errors in this PR?

tylerjereddy:
It reports tons of stuff, but I was planning to keep this PR pretty scoped for now.

I suspect some of the errors `mypy` reports may be related to pykokkos doing fancy things it does not understand, so my thought was that we'd likely want to scope down what we'd consider a genuine error at first.

NaderAlAwar (approved):

PR #17, opened 2022-04-01
ENH: improve reduction accumulator type flexibility

tylerjereddy:
* still a work in progress--this branch can compile
and run `02_simple_reduce.py` with the `pk.int64` type,
but somewhere in the control flow an overflow is
still happening for large values compared to NumPy

* `python runtests.py` passes 87 tests and fails 2
on this feature branch:
```
FAILED tests/test_lambdas.py::TestLambda::test_add - SystemExit: 1
FAILED tests/test_lambdas.py::TestLambda::test_add_squares - SystemExit: 1
```
- they look like type-related failures due to my crude initial
work here

TODO is probably something like:

- [ ] get some guidance from Nader on the general direction here
- [ ] write regression test vs. NumPy for the sum of squares reduction operation tolerating values that currently overflow with large integers instead of iterating with `rm -rf pk_cpp/ && python examples/kokkos/02_simple_reduce.py `

### DIFF
--- a/pykokkos/core/runtime.py
+++ b/pykokkos/core/runtime.py
@@ -4,7 +4,6 @@ from typing import Any, Callable, Dict, Optional, Tuple, Type, Union
 
 import numpy as np
 
-import kokkos
 from pykokkos.core.keywords import Keywords
 from pykokkos.core.translators import PyKokkosMembers
 from pykokkos.core.visitors import visitors_util
@@ -127,14 +126,6 @@ class Runtime:
 
         is_workunit_or_functor: bool = isinstance(entity, Callable)
         result = self.call_wrapper(entity, members, args, module, is_workunit_or_functor)
-
-        red_view_type = kokkos.double
-        for key, val in args.items():
-            if 'reduction_result_' in key:
-                # TODO: other types
-                if  "int64" in str(val):
-                    red_view_type = kokkos.int64
-        args["red_view_type"] = red_view_type
 
         if not is_workunit_or_functor:
             self.retrieve_results(entity, members, args)
@@ -282,9 +273,7 @@ class Runtime:
 
         for result in members.reduction_result_queue:
             name: str = f"reduction_result_{result}"
-            # TODO: need a switch that sets DataType based on
-            # kernel argument types/casting rules?
-            result_view = View([1], DataType.int64, MemorySpace.HostSpace)
+            result_view = View([1], DataType.double, MemorySpace.HostSpace)
             args[name] = result_view.array
 
         for result in members.timer_result_queue:
--- a/pykokkos/core/runtime.py
+++ b/pykokkos/core/runtime.py
@@ -4,6 +4,7 @@ from typing import Any, Callable, Dict, Optional, Tuple, Type, Union
 
 import numpy as np
 
+import kokkos
 from pykokkos.core.keywords import Keywords
 from pykokkos.core.translators import PyKokkosMembers
 from pykokkos.core.visitors import visitors_util
@@ -126,6 +127,14 @@ class Runtime:
 
         is_workunit_or_functor: bool = isinstance(entity, Callable)
         result = self.call_wrapper(entity, members, args, module, is_workunit_or_functor)
+
+        red_view_type = kokkos.double
+        for key, val in args.items():
+            if 'reduction_result_' in key:
+                # TODO: other types
+                if  "int64" in str(val):
+                    red_view_type = kokkos.int64
+        args["red_view_type"] = red_view_type
 
         if not is_workunit_or_functor:
             self.retrieve_results(entity, members, args)
@@ -273,7 +282,9 @@ class Runtime:
 
         for result in members.reduction_result_queue:
             name: str = f"reduction_result_{result}"
-            result_view = View([1], DataType.double, MemorySpace.HostSpace)
+            # TODO: need a switch that sets DataType based on
+            # kernel argument types/casting rules?
+            result_view = View([1], DataType.int64, MemorySpace.HostSpace)
             args[name] = result_view.array
 
         for result in members.timer_result_queue:
--- a/pykokkos/core/translators/bindings.py
+++ b/pykokkos/core/translators/bindings.py
@@ -2,7 +2,6 @@ import sys
 import re
 from typing import Dict, List, Optional, Tuple
 
-import pykokkos as pk
 from pykokkos.core import cppast
 from pykokkos.core.keywords import Keywords
 from pykokkos.core.visitors import cpp_view_type, KokkosMainVisitor, visitors_util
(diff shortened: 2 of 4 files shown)


pykokkos/core/runtime.py
```diff
@@ -4,6 +4,7 @@
 
 import numpy as np
 
+import kokkos
```
tylerjereddy, line 7: From reading through the code, I think there may be an alias or data structure I should use for type access instead of `kokkos` directly?

pykokkos/core/runtime.py
```diff
@@ -124,6 +125,13 @@ def execute(
         else:
             args["pk_kernel_name"] = name
 
+        red_view_type = kokkos.double
+        for key, val in args.items():
+            if 'reduction_result_' in key:
+                # TODO: other types
+                if  "int64" in str(val):
```
tylerjereddy, line 132: This feels pretty fragile--there must be something better than looking at the `repr` of the Kokkos objects for type analysis here.

pykokkos/core/translators/bindings.py
```diff
@@ -73,6 +73,29 @@ def get_kernel_params(
...
+        # integer type to use for the reduction return view
+        # type
+        if "int64_t" in params.values():
+            red_view_type = "int64_t"
+        elif "int32_t" in params.values():
+            # NOTE: which is preferred--preservation
+            # of the types in the reduction, or casting
+            # up to reduce likelihood of overflow?
+            red_view_type = "int32_t"
+    params["red_view_type"] = red_view_type
```
tylerjereddy, line 98: This feels pretty clunky..

pykokkos/core/runtime.py
NaderAlAwar: Yeah I've mostly been using `from pykokkos.bindings import kokkos`

pykokkos/core/runtime.py
NaderAlAwar: Would `isinstance()` work here?

NaderAlAwar:
This breaks for me if I add another field to `02_simple_reduce.py` which is not an integer type.

Is the intention here to have a different view for every reduction operation, and then  have the type of each view be based on the type of the accumulator?

tylerjereddy:
Still working on this locally--I need to write the tests carefully first, and have made a little progress there, and then probably revert a decent percentage of the stuff done here. So, not ready for additional review just yet.

pykokkos/core/translators/bindings.py
```diff
@@ -536,9 +537,23 @@ def bind_main_single(
...
+        if "pk.Acc" in element:
+            if "pk.int64" in element:
+                acc_type = "int64_t"
+            elif "pk.double" in element:
+                acc_type = "double"
+
     signature: str = generate_kernel_signature("void", kernel_name, params)
     instantiation: str = generate_functor_instance(functor, members)
-    acc: str = f"double {Keywords.Accumulator.value} = 0;"
+    acc: str = f"{acc_type} {Keywords.Accumulator.value} = 0;"
```
tylerjereddy, line 556: @NaderAlAwar The diff should be a bit cleaner to review now. AFAICT this change makes it possible mutate the `type` of the Accumulator, which it seems was previously hardcoded.

While the previous discussion focused on making it possible to change the reduction result view type to match the accumulator type, it seems that wouldn't have much value if the accumulator type is hardcoded anyway, so this is at least a first step that allows more of the new test cases that I wrote to pass compared with `develop`, but not all of them yet.

tests/test_parallelreduce.py
```diff
@@ -40,5 +44,53 @@ def test_add_squares(self):
...
+    def __init__(self, n):
+        self.N: int = n
+        self.total: pk.int64 = 0
+
+    @pk.main
+    def run(self):
+        self.total = pk.parallel_reduce(self.N, self.squaresum)
+
+    @pk.workunit
+    def squaresum(self, i: int, acc: pk.Acc[pk.int64]):
```
tylerjereddy, line 73: Locally I see only `1` test failure, for the largest integer value parametrized in the new test below. It appears to me that `i` is a 32-bit integer here, and that is causing overflow. Of course, I tried changing `i` to `pk.int64`, but this doesn't seem to translate or compile.

So, there appear to be a few different issues when attempting to match the NumPy reduction behavior here.

tests/test_parallelreduce.py
```diff
@@ -40,5 +44,53 @@ def test_add_squares(self):
...
+    # consistency with NumPy for lack
+    # of overflow, etc.
+    expected = np.sum(np.arange(series_max, dtype=dtype) ** 2)
+    if dtype == np.float64:
+        ss_instance = SquareSumFloat(series_max)
+    elif dtype == np.int64:
+        ss_instance = SquareSumInt(series_max)
+    pk.execute(pk.ExecutionSpace.OpenMP, ss_instance)
+    actual = ss_instance.total
+    assert_allclose(actual, expected)
```
tylerjereddy, line 93: This test exposes some overflow/typing issues that still remain as I note above, more than it really focuses on the reduction view type. Actually, I'm not sure we have a great handle for checking the reduction view type in the Python layer, since we only get the reduced `self.total` scalar back?

Nonetheless, I suspect we may agree that this test should be possible to make pass eventually.

tests/test_parallelreduce.py
```diff
@@ -40,5 +44,53 @@ def test_add_squares(self):
...
+    def run(self):
+        self.total = pk.parallel_reduce(self.N, self.squaresum)
+
+    @pk.workunit
+    def squaresum(self, i: float, acc: pk.Acc[pk.double]):
+        acc += i * i
+
+
+@pk.workload
+class SquareSumInt:
```
tylerjereddy, line 63: Is there a way for me to "parametrize" the workloads by type rather than copy/pasting and using slightly different types for testing like this? Seems tricky maybe because the type hints can't really be substituted in?

pykokkos/core/translators/bindings.py
```diff
@@ -2,6 +2,7 @@
 import re
 from typing import Dict, List, Optional, Tuple
 
+import pykokkos as pk
```
tylerjereddy, line 5: this can be removed I think

pykokkos/core/translators/bindings.py
```diff
@@ -536,9 +537,23 @@ def bind_main_single(
 
     main: List[str] = translate_mains(source, functor, members, pk_import)
     params: Dict[str, str] = get_kernel_params(members, False, True, real)
+
+    # fall back to the old hard-coded default
+    # for now--this includes cases where an
+    # accumulator is not even defined
+    acc_type = "double"
```
tylerjereddy, line 544: to clarify, this was needed for the `lambda` tests

tylerjereddy:
The tests pass locally, hopefully CI agrees.

This PR has basically been simplified to meet 1 purpose: improve the parallel reduction accumulator type flexibility beyond being hardcoded to `double`.

NaderAlAwar (approved):

PR #18, opened 2022-04-05
WIP, CI: Add initial CI with mypy

nawtrey:
### Description
* Add config file for `mypy` such that no errors are raised

* Add GitHub actions CI testing

This is an initial trial that I tested on my [fork](https://github.com/nawtrey/pykokkos/pull/1) with Python 3.8-3.10, which passed for all 3 Python versions. The CI is currently pinned to the `develop` branch, and I have the `mypy.ini` set to ignore basically every module. There are simpler ways of accomplishing this, but I went with a per-module approach so they can easily be removed one at a time, and corrected in subsequent PR's. Eventually it would be nice to have all of the modules passing, of course. 

Commenting out any module will result in some number of errors (as many as 50+ for some), with the exception of the `mypy-pykokkos.bindings.bindings.libpykokkos`, which only raised an error in the CI:
```
pykokkos/bindings/bindings/__init__.py:59: error: Skipping analyzing "pykokkos.bindings.bindings.libpykokkos": module is installed, but missing library stubs or py.typed marker  [import]
pykokkos/bindings/bindings/__init__.py:59: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 55 source files)
```

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -1,0 +1,33 @@
+name: Python Testing
+
+on:
+  push:
+    branches:
+      - develop
+  pull_request:
+    branches:
+      - develop
+
+jobs:
+  test_pykokkos:
+    strategy:
+      matrix:
+        platform: [ubuntu-latest]
+        python-version: ["3.8", "3.9", "3.10"]
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


nawtrey:
I don't think this will run unless the initial pipeline run is approved. I remember it being tricky to find but sometimes it appears at the bottom of the PR.

NaderAlAwar (approved):
Thanks Nik! I'll work on fixing some of the mypy errors

PR #20, opened 2022-04-09
WIP, CI: run tests in CI

tylerjereddy:
* run the actual tests in the CI--I developed this
locally using the [`act`](https://github.com/nektos/act) tool--the current blocker
seems to be that we have unguarded `import cupy`
in the both the codebase and the testsuite, so I'm
not sure how we do a pure `OpenMP` toolchain/testing
for vanilla GitHub Actions virtual machines that lack
GPUs/CUDA?

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -24,10 +24,21 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade numpy mypy
+          python -m pip install --upgrade numpy mypy cmake pytest pybind11
+      - name: Install pykokkos-base
+        run: |
+          cd /tmp
+          git clone https://github.com/kokkos/pykokkos-base.git
+          cd pykokkos-base
+          mkdir build && cd build
+          cmake -DENABLE_LAYOUTS=ON -DENABLE_MEMORY_TRAITS=OFF ../
+          make -j
       - name: Install pykokkos
         run: |
           python -m pip install .
       - name: mypy check
         run: |
           mypy pykokkos
+      - name: run tests
+        run: |
+          python runtests.py
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -24,15 +24,13 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade numpy mypy cmake pytest pybind11
+          python -m pip install --upgrade numpy mypy cmake pytest pybind11 scikit-build patchelf
       - name: Install pykokkos-base
         run: |
           cd /tmp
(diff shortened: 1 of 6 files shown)


.github/workflows/main_ci.yml
```diff
@@ -24,10 +24,21 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade numpy mypy
+          python -m pip install --upgrade numpy mypy cmake pytest pybind11
+      - name: Install pykokkos-base
```
tylerjereddy, line 28: Looks like this compilation completes in about 3 minutes or less in the Linux GitHub actions here.

.github/workflows/main_ci.yml
```diff
@@ -24,10 +24,21 @@ jobs:
...
+          make -j
       - name: Install pykokkos
         run: |
           python -m pip install .
       - name: mypy check
         run: |
           mypy pykokkos
+      - name: run tests
+        run: |
+          python runtests.py
```
tylerjereddy, line 44: The fact that the GitHub Actions jobs "pass" likely reveals an issue with `runtests.py`--I suspect it is not properly returning a non-zero exit code when `pytest` fails, so we should probably fix that--either while working on this PR or separately, since it would be important to see the CI fail when tests fail of course.

tylerjereddy: Ok, I pushed in a revision to `runtests.py` to address this--looks like the exit code is properly propagated to a CI failure caused by the testsuite now.

.github/workflows/main_ci.yml
NaderAlAwar: That's reasonable

.github/workflows/main_ci.yml
NaderAlAwar: looks good, thanks!

NaderAlAwar:
I guess we need another PR to fix the `import cupy` statements first?

tylerjereddy:
Ok, we're much closer now--I just pushed the `cupy` shims in here and 1 test is failing, but that should be quick floating point comparison test fix I'll try to adjust now:

```
=================================== FAILURES ===================================
____________________ TestASTTranslator.test_math_constants _____________________

self = <tests.test_AST_translator.TestASTTranslator testMethod=test_math_constants>

    def test_math_constants(self):
        expected_result: float = self.threads * (math.pi + math.e + math.tau)
        result: float = pk.parallel_reduce(self.range_policy, self.functor.math_constants)
    
>       self.assertEqual(expected_result, result)
E       AssertionError: 607.15[29](https://github.com/kokkos/pykokkos/runs/5978300833?check_suite_focus=true#step:8:29)894614212 != 607.1529894614213

tests/test_AST_translator.py:170: AssertionError
```

NaderAlAwar (approved):

NaderAlAwar:
Looks good, thanks!

Issue #22, opened 2022-04-15
TST: pytest.mark.parametrize support


tylerjereddy:
I suspect this is related to the top-level namespace/global namespace expectations in `pykokkos`.

The first test passes, the second does not. For now, probably low priority because I can use old-school approach. Good to bookmark it though.

```diff
diff --git a/tests/test_parallelreduce.py b/tests/test_parallelreduce.py
index bb8ccd9..3fb7633 100644
--- a/tests/test_parallelreduce.py
+++ b/tests/test_parallelreduce.py
@@ -2,6 +2,10 @@ import unittest
 
 import pykokkos as pk
 
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
+
 
 # Tests for correctness of pk.parallel_reduce
 @pk.functor
@@ -39,6 +43,28 @@ class TestParallelReduce(unittest.TestCase):
 
         self.assertEqual(expected_result, result)
 
+@pk.workload
+class SquareSum:
+    def __init__(self, n):
+        self.N: int = n
+        self.total: pk.double = 0
+
+    @pk.main
+    def run(self):
+        self.total = pk.parallel_reduce(self.N, self.squaresum)
+
+    @pk.workunit
+    def squaresum(self, i: int, acc: pk.Acc[pk.double]):
+        acc += i * i
+
+
+def test_squaresum_types():
+    series_max = 10
+    expected = np.sum(np.arange(series_max) ** 2)
+    ss_instance = SquareSum(series_max)
+    pk.execute(pk.ExecutionSpace.OpenMP, ss_instance)
+    actual = ss_instance.total
+    assert_allclose(actual, expected)
 
 if __name__ == '__main__':
     unittest.main()
```

```diff
diff --git a/tests/test_parallelreduce.py b/tests/test_parallelreduce.py
index bb8ccd9..09620c0 100644
--- a/tests/test_parallelreduce.py
+++ b/tests/test_parallelreduce.py
@@ -2,6 +2,10 @@ import unittest
 
 import pykokkos as pk
 
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
+
 
 # Tests for correctness of pk.parallel_reduce
 @pk.functor
@@ -39,6 +43,28 @@ class TestParallelReduce(unittest.TestCase):
 
         self.assertEqual(expected_result, result)
 
+@pk.workload
+class SquareSum:
+    def __init__(self, n):
+        self.N: int = n
+        self.total: pk.double = 0
+
[...]

PR #23, opened 2022-04-16
Parser: fix crash caused by functions decorated with chained names 

NaderAlAwar:
Fix parser crashing issue reported in kokkos/pykokkos#22.


tylerjereddy (approved):
Works with parametrized `pytest` tests for me locally as well, so let's put it in.

Issue #27, opened 2022-04-29
ENH: support full range of NumPy ufunc ops


tylerjereddy:
My initial testing is actually looking pretty positive for `pykokkos` providing fairly-easy access to feature-parity with the NumPy `ufuncs`. One case where it looks like we currently would need to write manual code is for `np.sign` (see below). I wonder if, given that we can use things like `exp` and `sqrt` as-is "just fine" with `pykokkos` (which is great!) if we might benefit from making a convenience wrapper for `sign` (and any other simple missing functions) as well?

```python
@pk.workload
class SqrtView1D:
    def __init__(self, threads: int, view: pk.View1D[pk.double]):
        self.threads: int = threads
        self.view: pk.View1D[pk.double] = view

    @pk.main
    def run(self) -> None:
        pk.parallel_for(self.threads, self.pfor)

    @pk.workunit
    def pfor(self, tid: int) -> None:
        self.view[tid] = sign(self.view[tid])


def test_1d_ufunc_vs_numpy():
    view: pk.View1d = pk.View([10], pk.double)
    view[:] = np.arange(10, dtype=np.float64)
    sqrt_kokkos_instance = SqrtView1D(threads=10, view=view)
    pk.execute(pk.ExecutionSpace.Default, sqrt_kokkos_instance)
    actual = sqrt_kokkos_instance.view
    expected = np.sign(range(10))
    assert_allclose(actual, expected)
```

```
File "/vast/home/treddy/github_projects/pykokkos/tests/test_views.py", line 335:
 undefined symbol sign used
```

NaderAlAwar:
Yeah this would be great. The reason we support the current math functions (like `exp` and `sqrt`) is because they can be trivially mapped to C++ math functions in `cmath`. Is there an equivalent C++ function for `np.sign`? If not, we could potentially define our own implementation of `np.sign` in C++ during translation.

Also, what other math functions should we support?

tylerjereddy:
> Also, what other math functions should we support?

I added a few comments about that in gh-28, which drafts in a comparison test. Some things are available in i.e., boost, but a lot of cases you can write in 1-2 lines of C++ like code, so not sure we gain too much in those cases.

The main thing this got me thinking more deeply about is whether you'd want to allow devs/users to avoid writing workloads completely by providing an actual "ufunc" interface in pykokkos--so something like `pk.sqrt(view)` to match `np.sqrt(array)`. It might help adoption, but not sure that's really on the menu in the near-term.

[...]

PR #28, opened 2022-04-30
TST: unary ufunc math test compare

tylerjereddy:
Related to gh-27

* add a test comparing pykokkos vs. numpy for
unary math ufuncs: https://numpy.org/doc/stable/reference/ufuncs.html#math-operations

* for cases where NumPy and C++ share the same "operational"
keyword, this seems to work well enough

* I've placed an `xfail` on the other cases

* that said, I'm not sure if we actually get much benefit
from providing a convenience function for something like
`np.reciprocal` vs. typing `1 / self.view[tid]` and `cbrt`
vs. i.e., using `pow()`

* note that Boost and the C complex headers may help
fill in some gaps here, but not sure we'd want to go
that route or not

* one thing this does make me thing about though is whether
pykokkos should consider offering i.e., `pk.sqrt()` type
stuff--kind of like the Pykokkos flavor of ufuncs for devs
that don't want to write out their own workloads for those
cases?


NaderAlAwar:
The tests look good, but do you think we should maybe move them to a new file? `test_ufuncs.py` or something?

I don't think we should add boost as a dependency. I'm also not too familiar with the C complex headers. What functions are available there?

> one thing this does make me thing about though is whether
pykokkos should consider offering i.e., pk.sqrt() type
stuff--kind of like the Pykokkos flavor of ufuncs for devs
that don't want to write out their own workloads for those
cases?

Yeah I think this would be a good idea.

tylerjereddy:
@NaderAlAwar I moved the tests to a new module as you suggested.

tests/test_ufuncs.py
```diff
@@ -0,0 +1,293 @@
...
+])
+def test_1d_unary_ufunc_vs_numpy(kokkos_test_class, numpy_ufunc):
+    # verify that we can easily recreate the functionality
+    # of most NumPy "unary" ufuncs on 1D views/arrays without much
+    # custom code
+    # NOTE: maybe we directly provide i.e., pk.sqrt(view)
+    # "pykokkos ufuncs" some day?
+    view: pk.View1d = pk.View([10], pk.double)
+    view[:] = np.arange(10, dtype=np.float64)
+    sqrt_kokkos_instance = kokkos_test_class(threads=10, view=view)
```
tylerjereddy, line 288: oh, I should change this name, it isn't just `sqrt` anymore

NaderAlAwar (approved):

PR #29, opened 2022-05-06
ENH: expose pk.reciprocal

tylerjereddy:
* draft of new `pk.reciprocal()` "ufunc" exposed
for initial usage with 1D arrays of type `double`

* the diff contains various comments/design questions,
which is why I've only drafted 1 ufunc here so far; another
thing to keep in mind is how closely we'd like to match
the arguments from the original ufunc:
https://numpy.org/doc/stable/reference/generated/numpy.reciprocal.html

It is a bit tricky to compare timings, but with 4 OpenMP threads set
in my environment, `pykokkos` can perform quite a bit faster after
paying the cost of the initial compilation (which NumPy paid
previously when it was built ahead of time):

<details>

<summary> Small Python timing script details are below the fold </summary>

```python
import time

import numpy as np
from numpy.testing import assert_allclose
import pykokkos as pk

def main():
    # set up data structures
    size = 100000000
    arr = np.arange(size, dtype=np.float64)
    view: pk.View1D[pk.double] = pk.View([size], pk.double)
    view[:] = arr[:]

    start_numpy = time.perf_counter()
    numpy_result = np.reciprocal(arr)
    end_numpy = time.perf_counter()

    start_pk = time.perf_counter()
    pk_result = pk.reciprocal(view)
    end_pk = time.perf_counter()

    assert_allclose(pk_result, numpy_result)

    print("Time (s) for NumPy:", end_numpy - start_numpy)
    print("Time (s) for pykokkos:", end_pk - start_pk)

if __name__ == "__main__":
    main()

```

</details>

First benchmark pass (includes compile):
```
INFO:root:translation 0.0007879569966462441
INFO:root:compilation 4.233883150998736
Time (s) for NumPy: 0.14787226800399367
Time (s) for pykokkos: 4.306444576002832
```

Second benchmark pass (cached `pk_cpp` folder):

```
Time (s) for NumPy: 0.14742803500121227
Time (s) for pykokkos: 0.0750463389995275
```

Also of note that is it would be slick if we could use `timeit` in `ipython` more easily with `pykokkos`, but there are several blockers there at the moment (`main` /global related scoping stuff I think), not to mention the caution needed for operating "in place" on  views when you do repeat calculations, so that's why I wrote my own script there for now.

Obviously, the performance improvement would likely be more drastic with larger arrays and using GPU, but even with just a few threads this is already faster than NumPy with a roughly similar level of code complexity (I think there's probably a way to initialize views with less awkwardness than I've done here?).

### DIFF
--- a/pykokkos/lib/ufuncs.py
+++ b/pykokkos/lib/ufuncs.py
@@ -1,33 +1,28 @@
+from functools import singledispatch
+
 import pykokkos as pk
 
 # NOTE: it might be slick if we could encapsulate
-# the workload code inside of the ufunc definitions,
+# the workunit code inside of the ufunc definitions,
 # but we currently require global scoping
 
 
-@pk.workload
-class ReciprocalView1D:
-    def __init__(self, view: pk.View1D[pk.double]):
-        # TODO: why doesn't view have a "size"
-        # attribute? it is basically the shape product
-        # in higher dims
-        self.threads: int = view.shape[0] # type: ignore
-        self.view: pk.View1D[pk.double] = view
+@pk.workunit
+def reciprocal_impl_1d_double(view: pk.View1D[pk.double], tid: int):
+    view[tid] = 1 / view[tid] # type: ignore
 
-    @pk.main
-    def run(self) -> None:
-        pk.parallel_for(self.threads, self.pfor)
+@pk.workunit
+def reciprocal_impl_1d_single(view: pk.View1D[pk.single], tid: int):
+    view[tid] = 1 / view[tid] # type: ignore
 
-    @pk.workunit
-    def pfor(self, tid: int) -> None:
-        self.view[tid] = 1 / self.view[tid] # type: ignore
 
 
 # TODO: how are we going to "ufunc dispatch" when the view
 # has a different type/precision? i.e., something other than
 # pk.double?
 
-def reciprocal(view: pk.View1D[pk.double]):
+@singledispatch
+def reciprocal(view):
     """
     Return the reciprocal of the argument, element-wise.
 
@@ -47,15 +42,16 @@ def reciprocal(view: pk.View1D[pk.double]):
         This function is not designed to work with integers.
 
     """
-    workload_instance = ReciprocalView1D(view=view)
-    # NOTE: should our ufuncs have an option to allow
-    # CUDA vs. OpenMP execution directly in their call
-    # signature?
-    pk.execute(pk.ExecutionSpace.Default, workload_instance)
-    # TODO: how should we design this?
-    # it seems a bit problematic to both mutate the data
-    # underneath a view in-place and return a reference to the view?
-    # returning a new view/"array" would be more similar to
-    # NumPy?
-    return workload_instance.view
+    pass
+
+
+@reciprocal.register
+def _(view: pk.View1D[pk.double]):
+    pk.parallel_for(view.shape[0], reciprocal_impl_1d_double, view=view)
+    return view
+
+@reciprocal.register
+def _(view: pk.View1D[pk.single]):
+    pk.parallel_for(view.shape[0], reciprocal_impl_1d_single, view=view)
+    return view
     
--- a/pykokkos/lib/ufuncs.py
+++ b/pykokkos/lib/ufuncs.py
@@ -1,39 +1,28 @@
-from functools import singledispatch
-
 import pykokkos as pk
 
-# NOTE: it might be slick if we could encapsulate
-# the workunit code inside of the ufunc definitions,
-# but we currently require global scoping
-
 
 @pk.workunit
-def reciprocal_impl_1d_double(view: pk.View1D[pk.double], tid: int):
-    view[tid] = 1 / view[tid] # type: ignore
-
-@pk.workunit
-def reciprocal_impl_1d_single(view: pk.View1D[pk.single], tid: int):
+def reciprocal_impl_1d_double(tid: int, view: pk.View1D[pk.double]):
     view[tid] = 1 / view[tid] # type: ignore
 
 
+@pk.workunit
+def reciprocal_impl_1d_float(tid: int, view: pk.View1D[pk.float]):
+    view[tid] = 1 / view[tid] # type: ignore
 
-# TODO: how are we going to "ufunc dispatch" when the view
-# has a different type/precision? i.e., something other than
-# pk.double?
 
-@singledispatch
 def reciprocal(view):
     """
     Return the reciprocal of the argument, element-wise.
 
     Parameters
     ----------
-    view : pk.View1D
+    view : pykokkos view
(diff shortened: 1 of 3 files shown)


pykokkos/lib/ufuncs.py
```diff
@@ -0,0 +1,61 @@
...
+        pk.parallel_for(self.threads, self.pfor)
+
+    @pk.workunit
+    def pfor(self, tid: int) -> None:
+        self.view[tid] = 1 / self.view[tid] # type: ignore
+
+
+# TODO: how are we going to "ufunc dispatch" when the view
+# has a different type/precision? i.e., something other than
+# pk.double?
```
tylerjereddy, line 28: what would you think about using Python single dispatch here: https://docs.python.org/3/library/functools.html#functools.singledispatch ?

pykokkos/lib/ufuncs.py
```diff
@@ -0,0 +1,61 @@
...
+# the workload code inside of the ufunc definitions,
+# but we currently require global scoping
+
+
+@pk.workload
+class ReciprocalView1D:
+    def __init__(self, view: pk.View1D[pk.double]):
+        # TODO: why doesn't view have a "size"
+        # attribute? it is basically the shape product
+        # in higher dims
```
NaderAlAwar, line 13: We should probably add this (in another PR)

pykokkos/lib/ufuncs.py
NaderAlAwar: I'm not too familiar with this. I looked through the docs quickly, it seems that it will look at the type annotations and call the right function based on the type at runtime? If so that seems good, I think we should use it.

NaderAlAwar: Of course, we'd still need to implement a different kernel for each view type

pykokkos/lib/ufuncs.py
```diff
@@ -0,0 +1,61 @@
+import pykokkos as pk
+
+# NOTE: it might be slick if we could encapsulate
+# the workload code inside of the ufunc definitions,
+# but we currently require global scoping
+
+
+@pk.workload
```
NaderAlAwar, line 8: I feel like we should just use a standalone workunit here instead of a workload. It reduces overhead slightly since we won't be creating an instance of the workload

tests/test_ufuncs.py
```diff
@@ -291,3 +291,17 @@ def test_1d_unary_ufunc_vs_numpy(kokkos_test_class, numpy_ufunc):
...
+        (pk.reciprocal, np.reciprocal),
+])
+def test_1d_exposed_ufuncs_vs_numpy(pk_ufunc, numpy_ufunc):
+    # test the ufuncs we have exposed in the pk namespace
+    # vs. their NumPy equivalents
+    expected = numpy_ufunc(np.arange(10, dtype=np.float64))
+
+    # TODO: parametrize test with more types somehow?
+    view: pk.View1d = pk.View([10], pk.double)
+    view[:] = np.arange(10, dtype=np.float64)
```
NaderAlAwar, line 305: you could do view.fill(10)

NaderAlAwar:
Looks good. I'll look into the ipython pykokkos issue.

tylerjereddy:
I made some revisions, but there is one substantial issue with using `singledispatch` as noted in the commit message:

    * switch to simpler workunits for the reciprocal
    ufunc implementation, and use `functools.singledispatch`
    for argument type dispatching
    
    * unfortunately, at the moment, it looks like `pykokkos`
    view type annotations are not suitable for `singledispatch`
    because they are not not formally Python classes;
    generic classes are mostly incompatible with runtime checks
    I think per: https://stackoverflow.com/a/58445388/2942522
    
    ```
    _________________________________________________ ERROR collecting tests/test_ufuncs.py _________________________________________________
    tests/test_ufuncs.py:1: in <module>
        import pykokkos as pk
    pykokkos/__init__.py:15: in <module>
        from pykokkos.lib.ufuncs import reciprocal # type: ignore
    pykokkos/lib/ufuncs.py:49: in <module>
        def _(view: pk.View1D[pk.double]):
    /usr/lib/python3.10/functools.py:873: in register
        raise TypeError(
    E   TypeError: Invalid annotation for 'view'. pykokkos.interface.views.View1D[pykokkos.interface.data_types.double] is not a class.
    ```
    
    - pretty sure that is a valid `view` type annotation right? this may
    require some plumbing changes?

pykokkos/lib/ufuncs.py
```diff
@@ -0,0 +1,57 @@
...
+    pass
+
+
+@reciprocal.register
+def _(view: pk.View1D[pk.double]):
+    pk.parallel_for(view.shape[0], reciprocal_impl_1d_double, view=view)
+    return view
+
+@reciprocal.register
+def _(view: pk.View1D[pk.single]):
```
tylerjereddy, line 54: well, maybe we could try `workunit` decoration here? I don't know, that seems awkward to have two `workunits` per type-kernel; we could also try moving this code into the `workunit` functions above, and decorating those direclty at the cost of some code duplication..

tylerjereddy: also not sure how double-decoration with `register` and `workunit` will behave

pykokkos/lib/ufuncs.py
```diff
@@ -1,33 +1,28 @@
...
-        self.view: pk.View1D[pk.double] = view
+@pk.workunit
+def reciprocal_impl_1d_double(view: pk.View1D[pk.double], tid: int):
+    view[tid] = 1 / view[tid] # type: ignore
 
-    @pk.main
-    def run(self) -> None:
-        pk.parallel_for(self.threads, self.pfor)
+@pk.workunit
+def reciprocal_impl_1d_single(view: pk.View1D[pk.single], tid: int):
```
NaderAlAwar, line 15: The type you're looking for is probably `pk.float`, `single` refers to a function

pykokkos/lib/ufuncs.py
NaderAlAwar: Not sure if a `workunit` decoration would work here, since we're not calling these functions directly, we're calling `parallel_for`

NaderAlAwar:
So the view type annotations that we use e.g. `pk.View1D[pk.float]` are not the actual type of the view. These are defined [here](https://github.com/kokkos/pykokkos/blob/8971f276f0aea445ba9bb48ef8d394ddd35d9815/pykokkos/interface/views.py#L466), and they're only used during translation to figure out the C++ type. It might be possible to get this approach to work, but it would require us to rework the `View` constructor.

tylerjereddy:
Ok, I've revised this to use a cruder approach to "type" dispatching and expanded the test to parametrize over single and double floating point precision, which seems to pass locally at least.

NaderAlAwar (approved):

Issue #32, opened 2022-05-19
BUG: test failures vs. kokkos develop branch


tylerjereddy:
Full test suite result vs. Kokkos `develop` branch (OpenMP only): ` = 66 failed, 31 passed, 9 skipped, 9 xfailed, 2 warnings in 148.22s (0:02:28) ==`

Most of the failures look like this one:

```
________________ TestKokkosFunctionsTranslator.test_nested_sum _________________
| 
| self = <tests.test_kokkosfunctions_translator.TestKokkosFunctionsTranslator testMethod=test_nested_sum>
| 
|     def test_nested_sum(self):
|         expected_result: int = self.threads * (self.i_1 + self.i_2)
| >       result: int = pk.parallel_reduce(self.range_policy, self.functor.add_nested)
| 
| tests/test_kokkosfunctions_translator.py:173: 
| _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
| pykokkos/interface/parallel_dispatch.py:220: in parallel_reduce
|     return reduce_body("reduce", *args, **kwargs)
| _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
| 
| operation = 'reduce'
| args = {'b_1': True, 'f_1': 5.5, 'f_2': 1.3, 'i_1': 7, ...}, kwargs = {}
| args_to_hash = ['add_nested', 'reduce'], args_not_to_hash = {}
| a = <bound method KokkosFunctionsTestReduceFunctor.add_nested of <tests.test_kokkosfunctions_translator.KokkosFunctionsTestReduceFunctor object at 0x7f36054e8520>>
| to_hash = frozenset({'add_nested', 'reduce'}), cache_key = -4624954395847863019
| func = <built-in method wrapper_add_nested of PyCapsule object at 0x7f3607e9a8e0>
| handled_args = HandledArgs(name=None, policy=<pykokkos.interface.execution_policy.RangePolicy object at 0x7f36054ea3e0>, workunit=<bo...est_kokkosfunctions_translator.KokkosFunctionsTestReduceFunctor object at 0x7f36054e8520>>, view=None, initial_value=0)
| 
|     def reduce_body(operation: str, *args, **kwargs) -> Union[float, int]:
[...]

PR #33, opened 2022-05-19
CI: nightly vs. Kokkos develop

### LINKED ISSUE #30: CI: test against Kokkos develop
The idea of having a CI entry that runs ~nightly with Kokkos `develop` branch was brought up by Damiean in a recent meeting--I think the Kokkos team is interested in having us make sure that the cross-project testing catches any major issues before releases go out.

I think this seems like an "ok" idea--main challenge for me would be to get `pykokkos-base` to use the correct Kokkos checkout--perhaps it is as simple as getting a submodule checked out at the correct hash, but maybe there's a bit more to it than that?

tylerjereddy:
Fixes #30

* add a nightly cron test of our `develop` vs. Kokkos `develop`
branch

* you can run/test it locally using the [`act`](https://github.com/nektos/act) tool
with i.e., `act -j test_vs_kokkos_develop`

* at the moment, it correctly detects the test failures
described in gh-32

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,59 @@
+name: Nightly_Testing_vs_Kokkos_develop
+
+on:
+  schedule:
+    # 5 am UTC is late evening for current dev
+    # team
+    - cron: '00 5 * * *'
+
+jobs:
+  test_vs_kokkos_develop:
+    strategy:
+      matrix:
+        platform: [ubuntu-latest]
+        python-version: ["3.10"]
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - uses: actions/checkout@v3
+      - run: git checkout develop
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install --upgrade numpy mypy cmake pytest pybind11 scikit-build patchelf ninja
+      - name: Build Kokkos develop branch
+        run: |
+          cd /tmp
+          mkdir kokkos_install
+          git clone https://github.com/kokkos/kokkos.git
+          cd kokkos
+          git checkout develop
+          mkdir build && cd build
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: mypy check
+        run: |
+          mypy pykokkos
+      - name: run tests
+        run: python runtests.py
+        # when you build pykokkos-base with an external Kokkos,
+        # it fails to copy the lib/ with Kokkos shared libs
+        # to the installed Python package dir (which it does in
+        # the submodule scenario), so set it manually
+        # export PK_KOKKOS_LIB_PATH=/tmp/kokkos_install/lib
```
tylerjereddy, line 57: @jrmadsen fyi

jrmadsen: Yeah, that is intentional, it is pretty uncommon to copy files from an external installation into your installation. Few options here based on what the problem is:

If the issue is just locating the libraries at runtime, try making sure that `-DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON` so that the pykokkos-base modules have an RPATH to the external Kokkos installation (i.e. so you don't have to append their path to the `LD_LIBRARY_PATH`). 

Another option is to tweak pykokkos-base to deduce the location of the installed Kokkos libraries in cmake (or set it to lib/ when kokkos is a submodule), e.g. `set(PK_KOKKOS_LIB_PATH /path/to/installed/kokkos/libs)` (where that path is deduced) and configure a file like this:

```cpp
namespace
{
// when the library gets loaded, this will set the env variable to the default path but zero as 
// last parameter means it will not override it if explicitly set
auto _set_pk_kokkos_lib_env = setenv("PK_KOKKOS_LIB_PATH", @PY_KOKKOS_LIB_PATH@, 0);
}

// function for getting the kokkos lib path in python
std::string get_kokkos_lib_path()
{
    // this will never be a nullptr as long at is NOT called during the static initialization
    static auto _v = std::string{ getenv("PK_KOKKOS_LIB_PATH") };
    return _v;
}

PYBIND11_MODULE(...)
{
     // create binding
     pk.def("get_kokkos_lib_path", &get_kokkos_lib_path, "Get path to kokkos-libraries");
}
```

PR #35, opened 2022-06-06
BUG: cache no dead objects

### LINKED ISSUE #34: BUG: thread/reference safety for views or kernels?
I was in the middle of typing up a more detailed description of the problem when my power went out, but I'll try to provide some more concise info for my issue here from what I can do on my laptop.

Working on the `develop` branch, if we simplify `tests/test_ufuncs.py` to only contain this while debugging: 

```python
import pykokkos as pk

import numpy as np
from numpy.testing import assert_allclose
import pytest

@pytest.mark.parametrize("pk_ufunc, numpy_ufunc", [
        (pk.reciprocal, np.reciprocal),
])
@pytest.mark.parametrize("pk_dtype, numpy_dtype", [
        (pk.double, np.float64),
        (pk.float, np.float32),
])
def test_1d_exposed_ufuncs_vs_numpy(pk_ufunc,
                                    numpy_ufunc,
                                    pk_dtype,
                                    numpy_dtype):
    # test the ufuncs we have exposed in the pk namespace
    # vs. their NumPy equivalents
    expected = numpy_ufunc(np.arange(10, dtype=numpy_dtype))

    view: pk.View1d = pk.View([10], pk_dtype)
    view[:] = np.arange(10, dtype=numpy_dtype)
    actual = pk_ufunc(view=view)
    assert_allclose(actual, expected)
```

The test will pass just fine as-is with `python runtests.py -t tests/test_ufuncs.py`. But, if we substantially increase the number of repeats like this (example only, I will use far more repeats in my testing):

```diff
 @pytest.mark.parametrize("pk_ufunc, numpy_ufunc", [
         (pk.reciprocal, np.reciprocal),
 ])
 @pytest.mark.parametrize("pk_dtype, numpy_dtype", [
         (pk.double, np.float64),
         (pk.float, np.float32),
+        (pk.float, np.float32),
+        (pk.float, np.float32),
+        (pk.float, np.float32),
+        (pk.float, np.float32),
+        (pk.float, np.float32),
 ])
 def test_1d_exposed_ufuncs_vs_numpy(pk_ufunc,
                                     numpy_ufunc,
```

I will start getting non-deterministic test failure ratios:

- `35 failed, 667 passed`
- `21 failed, 681 passed`
- `17 failed, 685 passed`

This seems bad--the kernel/workunit for `reciprocal` is incredibly simple, so my initial suspicion is something related to kernel/view thread safety maybe? 

The nature of the failure seems to be a somewhat-stochastic restoration of the original view values:

```
E       
E       x and y +inf location mismatch:
E        x: array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.], dtype=float32)
E        y: array([     inf, 1.      , 0.5     , 0.333333, 0.25    , 0.2     ,
E              0.166667, 0.142857, 0.125   , 0.111111], dtype=float32)
```

This is a simplified reproducer of a more complex situation that arose while I was adding more tests for the new `sqrt` `pykokkos` ufunc in feature branch `treddy_more_1d_ufuncs`. In that case, separate parametrized pytest tests were interfering with each other in bizarre ways by resetting the values in views to their originals. I also had higher failure ratios for the `sqrt` ufunc with high repeat counts. 

So, my first question is--is this a known issue? I can reproduce on two separate machines at least.

tylerjereddy:
Fixes #34

* ban the retrieval of objects from the cache
when any object in the cache has a reference count
of `0`, since this likely means that the reference
to a `pykokkos` object has been replaced by another
object, and the cache probably holds something invalid
at that point

* add a regression test, `test_caching()` that has a sufficiently
large number of iterations to trigger the Python garbage collection
machinery, and trigger this tricky issue; when it succeeds, this
test requires about 16 seconds locally

* I've added some comments to the source because we may need
to revisit the genuine value of the cache at some point, given
that views are mutable and not really safe to cache the data
they point to?

### DIFF
--- a/pykokkos/interface/parallel_dispatch.py
+++ b/pykokkos/interface/parallel_dispatch.py
@@ -1,3 +1,4 @@
+import gc
 from dataclasses import dataclass
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
@@ -142,10 +143,25 @@ def parallel_for(*args, **kwargs) -> None:
     cache_key: int = hash(to_hash)
 
     if cache_key in workunit_cache:
-        func, args = workunit_cache[cache_key]
-        args.update(args_not_to_hash)
-        func(**args)
-        return
+        dead_obj = 0
+        func, newargs = workunit_cache[cache_key]
+        for arg in newargs.values():
+            # see gh-34
+            # reject cache retrieval when an object in the
+            # cache has a reference count of 0 (presumably
+            # only possible because of the C++/pybind11 infra;
+            # normally a refcount of 1 is the lowest for pure
+            # Python objects)
+            # NOTE: is the cache genuinely useful now though?
+            ref_count = len(gc.get_referrers(arg))
+            if ref_count == 0:
+                dead_obj += 1
+                break
+        if not dead_obj:
+            args = newargs
+            args.update(args_not_to_hash)
+            func(**args)
+            return
 
     handled_args: HandledArgs = handle_args(True, args)
     func, args = runtime_singleton.runtime.run_workunit(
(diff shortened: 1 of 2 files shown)


tylerjereddy:
This was superseded by (included in) gh-36, so closing.

PR #36, opened 2022-06-12
ENH: add pk.sqrt ufunc in 1D

tylerjereddy:
* add `double` and `float` versions of `pk.sqrt()`, and some tests to enforce behavioral similarity with `np.sqrt()` in 1D
* this PR temporarily cherry-picks the commit from gh-35, so that we can avoid caching/reference management issues that would otherwise prevent the expanded ufunc tests from passing

### DIFF
--- a/pykokkos/__init__.py
+++ b/pykokkos/__init__.py
@@ -16,7 +16,8 @@ from pykokkos.lib.ufuncs import (reciprocal, # type: ignore
                                  log,
                                  log2,
                                  log10,
-                                 log1p)
+                                 log1p,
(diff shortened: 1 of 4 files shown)


NaderAlAwar (approved):
We still need to benchmark this, but I think the cache is still preferable to going through the pykokkos runtime

NaderAlAwar:
Could you resolve the conflicts? I think I merged the PRs in the wrong order

tylerjereddy:
This has an approval from Nader and CI is passing so I'll go ahead and self-merge.

PR #39, opened 2022-06-15
ENH: 2D reciprocal ufunc

tylerjereddy:
* `pk.reciprocal` ufunc now handles
2D views and a test against NumPy 2D
behavior has been added

### DIFF
--- a/pykokkos/lib/ufuncs.py
+++ b/pykokkos/lib/ufuncs.py
@@ -11,6 +11,18 @@ def reciprocal_impl_1d_float(tid: int, view: pk.View1D[pk.float]):
     view[tid] = 1 / view[tid] # type: ignore
 
 
+@pk.workunit
+def reciprocal_impl_2d_double(tid: int, view: pk.View2D[pk.double]):
+    for i in range(view.extent(1)): # type: ignore
+        view[tid][i] = 1 / view[tid][i] # type: ignore
+
+
+@pk.workunit
+def reciprocal_impl_2d_float(tid: int, view: pk.View2D[pk.float]):
+    for i in range(view.extent(1)): # type: ignore
+        view[tid][i] = 1 / view[tid][i] # type: ignore
+
+
 def reciprocal(view):
     """
     Return the reciprocal of the argument, element-wise.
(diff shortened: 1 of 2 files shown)


pykokkos/lib/ufuncs.py
```diff
@@ -11,6 +11,18 @@ def reciprocal_impl_1d_float(tid: int, view: pk.View1D[pk.float]):
...
+@pk.workunit
+def reciprocal_impl_2d_double(tid: int, view: pk.View2D[pk.double]):
+    for i in range(view.extent(1)): # type: ignore
+        view[tid][i] = 1 / view[tid][i] # type: ignore
+
+
+@pk.workunit
+def reciprocal_impl_2d_float(tid: int, view: pk.View2D[pk.float]):
+    for i in range(view.extent(1)): # type: ignore
+        view[tid][i] = 1 / view[tid][i] # type: ignore
```
tylerjereddy, line 23: Is this still performant when `view.extent(1) >> view.extent(0)`? Is there a way to write this more efficiently in terms of using parallel units/distributing work automatically?

tylerjereddy:
Digging into the 2D performance/thread iteration issues a bit farther with a simple bench script:

```python
from timeit import default_timer as timer

import pykokkos as pk

def bench():
    v = pk.View([2, int(1e6)], pk.double)
    v.fill(5)
    pk.reciprocal(v)


if __name__ == "__main__":
    for i in range(10):
        start = timer()
        bench()
        end = timer()
        print("Time (s):", end - start)
```

And varying the number of threads with incantations like this: `rm -rf pk_cpp && OMP_NUM_THREADS=1 python test.py`
on a machine with 10 physical cores/20 threads:

test_2_dims


vs. the 1D case where I thought we might get better splitting of the work between threads?

```python
def bench():
    v = pk.View([int(1e6)], pk.double)
    v.fill(5)
    pk.reciprocal(v)
```

test_1_dims

Neither case really seems to scale that well. Maybe I need much larger data in the benchmarks and/or randomized data as well.

tylerjereddy:
If we scale up to 1 billion columns (see below), the performance scaling looks decent in  both 1D and 2D, so perhaps I actually don't need to fret too much here about the details of 2D/multi-dimensional iteration? In other words, I've perhaps made incorrect assumptions about what the assignment of `tid` to the left-most index actually means with respect to "teams" of threads, and the work that happens across columns.

test_1_dims
test_2_dims

NaderAlAwar (approved):

PR #40, opened 2022-06-30
BLD: use PEP518 for pip install

tylerjereddy:
* modernize our `pip install` by using the [PEP518](https://peps.python.org/pep-0518/)
`pyproject.toml` configuration, which basically switches
us to the modern process where `pip` will first build
a wheel in an isolated env and then install that binary wheel
into the current user env

* note that unfortunately this doesn't really make dealing
with the friction of needing the complex `pykokkos-base`
package any easier, as I describe in
https://github.com/kokkos/pykokkos-base/issues/41

* we don't actually need `pykokkos-base` to *build*
`pykokkos`, so much like a user needs to provide a suitable
version of NumPy when working with SciPy, we need the user
to provide a suitable version of `pykokkos-base` *separately*
from the build/install of `pykokkos`; I suspect the best
way to make this easier would be to provide `pykokkos-base`
wheels on PyPI, which I don't see at the moment (likely a substantial
lift, but could be worth it)

### DIFF
--- a/pykokkos/core/run_debug.py
+++ b/pykokkos/core/run_debug.py
@@ -62,7 +62,7 @@ def call_workunit(
     workunit: Callable[..., None],
     index: Union[int, Tuple[int, int], TeamMember],
     acc: Acc,
-    **kwargs
+    **kwargs,
 ) -> None:
     """
     Run a workunit for a single iteration
@@ -99,7 +99,7 @@ def run_workunit_debug(
     workunit: Callable[..., None],
     operation: str,
     initial_value=0,
-    **kwargs
+    **kwargs,
 ) -> Optional[Union[float, int]]:
     """
     Run a workunit in Python


tylerjereddy:
I should add, this does make it slightly easier for us to provide wheels for `pykokkos` if we want to in the future, though building `pykokkos` from source is fairly easy at the moment anyway.

pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[build-system]
+requires = [
+    "setuptools<60.0.0",
+    "wheel",
+]
+
+[project]
+dependencies = [
+    "numpy>=1.19.5",
+    "pykokkos-base",
```
tylerjereddy, line 10: Note that with `pip` these don't get installed in the user environment automatically, it is basically just metadata about what we need at runtime. Indeed, the idea is to only provide `pykokkos` into the user env without mutating it in any other way. I believe some other install tools may be able to use this data, though I haven't use them. The NumPy version is based loosely on [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html).

pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[build-system]
+requires = [
+    "setuptools<60.0.0",
+    "wheel",
+]
```
tylerjereddy, line 5: We really need very little to install `pykokkos`, and it is lightning fast, so no huge advantage to wheels just yet I suppose.

pyproject.toml
NaderAlAwar: I see, so there's no way to install pykokkos + pykokkos-base with a single pip command?

tylerjereddy: You could add `pykokkos-base` to the `requires` section above and then do i.e., `pip install --no-build-isolation`, but `requires` is really only supposed to be used for genuine build *requirements*.

It still seems to me like the real problem is that `pykokkos-base` is a pain to install for people outside the project devs.

tylerjereddy:
Since I guess this was four years old, in a follow up it may be worth considering increasing the lower bound of `numpy>=1.19.5` in `pyproject.toml`. If you set it to `>=2.0.0` the wheel will still be backwards compatible with the `1.26.x` series, so that's popular choice.

PR #52, opened 2022-07-19
ENH: add sum ufunc up to 3D

tylerjereddy:
* add 1D-3D support for the `sum()` reduction ufunc;
this is likely the first example of a reduction ufunc
I've added so far

* satisfying the `mypy` type hint analysis caused some
tests to fail, so I'm suppressing the type hint checks
for the `ufuncs` module for now

* most comments/questions are in the source

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -92,3 +92,5 @@ ignore_errors = True
 [mypy-pykokkos.core.translators.bindings]
 ignore_errors = True
 
+[mypy-pykokkos.lib.ufuncs]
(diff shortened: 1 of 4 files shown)


tylerjereddy:
@NaderAlAwar I think I've dealt with the merge conflicts here and CI is passing, so maybe this is the next one to do code review for.

tylerjereddy:
formally, I don't think `sum()` is considered a `ufunc`, perhaps b/c a reduction op

Issue #54, opened 2022-07-20
ENH, MAINT: how to handle other array-like for ufuncs?


tylerjereddy:
NumPy ufuncs can generally accept array-like objects such as pure Python lists and even scalars. To provide a similar API in pykokkos, we'd need a way to coerce the input, perhaps via `np.asarray()` or similar, though some thought may be needed for cases where we want to send the data directly to `pykokkos` without needing to path through NumPy.

I noticed this when trying to replace `np.cumsum()` with `pk.cumsum()` in the SciPy source file `scipy/spatial/_spherical_voronoi.py`, with a traceback like this:

```
scipy/spatial/tests/test_spherical_voronoi.py:325: in test_equal_area_reconstitution
    areas = sv.calculate_areas()
        dim        = 3
        n          = 12
        points     = array([[ 0.        , -0.52573111, -0.85065081],
       [ 0.        , -0.52573111,  0.85065081],
       [ 0.        ,  ...5065081,  0.        ],
       [ 0.52573111, -0.85065081,  0.        ],
       [ 0.52573111,  0.85065081,  0.        ]])
        poly       = 'icosahedron'
        self       = <scipy.spatial.tests.test_spherical_voronoi.TestSphericalVoronoi object at 0x7f686edc6320>
        sv         = <scipy.spatial._spherical_voronoi.SphericalVoronoi object at 0x7f686cc6c610>
scipy/spatial/_spherical_voronoi.py:340: in calculate_areas
    return self._calculate_areas_3d()
        self       = <scipy.spatial._spherical_voronoi.SphericalVoronoi object at 0x7f686cc6c610>
scipy/spatial/_spherical_voronoi.py:266: in _calculate_areas_3d
    csizes = pk.cumsum(sizes)
        self       = <scipy.spatial._spherical_voronoi.SphericalVoronoi object at 0x7f686cc6c610>
        sizes      = [5, 5, 5, 5, 5, 5, ...]
../../../../../pykokkos/pykokkos/lib/ufuncs.py:421: in cumsum
    range_policy = pk.RangePolicy(pk.ExecutionSpace.Default, 0, view.shape[0])
E   AttributeError: 'list' object has no attribute 'shape'
        arr_type   = 'kokkos'
        view       = [5, 5, 5, 5, 5, 5, ...]

```

jrmadsen:
You just need to wrap the list into a object which satisfies the buffer protocol: https://docs.python.org/3/c-api/buffer.html. 

[...]

PR #55, opened 2022-07-20
ENH: pk.cumsum() ufunc

tylerjereddy:
* implement `pk.cumsum()` and test it against the NumPy
implementation; I think this is the first implementation
of a "parallel scan" ufunc that I've added in

* support is added for these types initially: `int32`,
`int64`, `float`, and `double`; supported dimensions
on views and NumPy arrays are 1 to 3 inclusive

* most notes/comments are in the code changes here

* ignore mypy type checking for `ufuncs` module for now,
as done in other recent PRs..

* note that practical usage is still somewhat hampered by
gh-54, because i.e., SciPy often passes in a list or other
"array-like" into a ufunc instead of an array/view directly

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -92,3 +92,5 @@ ignore_errors = True
 [mypy-pykokkos.core.translators.bindings]
 ignore_errors = True
 
+[mypy-pykokkos.lib.ufuncs]
+ignore_errors = True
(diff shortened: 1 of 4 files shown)


IvanGrigorik:
I'm not sure what's wrong with this workload, and if this is the problem with the code or the GitHub runner. I tried to run this locally, and everything works
Looks like runner overloads because of this pytest. I will try to isolate it.

Issue #56, opened 2022-07-25
MAINT, ENH: passing the array api tests


tylerjereddy:
I'm not sure if we actually want/need to pass these tests, but for the array API tests:
- https://github.com/data-apis/array-api-tests
- https://data-apis.org/array-api/2021.12/verification_test_suite.html

If I run:
- `export ARRAY_API_TESTS_MODULE=pykokkos`
- `pytest array_api_tests/`

I get the errors below the fold:

<details>

```
================================================================================================================================ test session starts =================================================================================================================================
platform linux -- Python 3.10.4, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/tyler/github_projects/array-api-tests
plugins: forked-1.4.0, hypothesis-6.53.0, xdist-2.5.0
collected 3116 items / 7 errors                                                                                                                                                                                                                                                      

======================================================================================================================================= ERRORS =======================================================================================================================================
_______________________________________________________________________________________________________________ ERROR collecting array_api_tests/test_array_object.py ________________________________________________________________________________________________________________
array_api_tests/test_array_object.py:15: in <module>
    from .test_operators_and_elementwise_functions import oneway_promotable_dtypes
<frozen importlib._bootstrap>:1027: in _find_and_load
    ???
<frozen importlib._bootstrap>:1006: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:688: in _load_unlocked
    ???
../../python_310_pykokkos_work/lib/python3.10/site-packages/_pytest/assertion/rewrite.py:168: in exec_module
    exec(co, module.__dict__)
array_api_tests/test_operators_and_elementwise_functions.py:611: in <module>
    @given(*hh.two_mutual_arrays(dh.float_dtypes))
array_api_tests/hypothesis_helpers.py:385: in two_mutual_arrays
    assert len(dtypes) > 0  # sanity check
E   AssertionError
____________________________________________________________________________________________________________ ERROR collecting array_api_tests/test_creation_functions.py _____________________________________________________________________________________________________________
array_api_tests/test_creation_functions.py:15: in <module>
    from .test_operators_and_elementwise_functions import oneway_promotable_dtypes
<frozen importlib._bootstrap>:1027: in _find_and_load
    ???
<frozen importlib._bootstrap>:1006: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:688: in _load_unlocked
    ???
[...]

PR #57, opened 2022-08-09
ENH: array API start

tylerjereddy:
Related to #56. This at least allows us to run the array API conformance test suite, rather than erroring out at import time. I suspect some of the changes here we won't really want to make, but on the flip side this gives folks an idea of some of the changes I needed to get us even remotely close to being able to run the suite.

* support the bare minimum required types per:
https://data-apis.org/array-api/2021.12/API_specification/data_types.html
(otherwise, we can't even import the array API tests, let
alone run them--for now, some types are basically just "stubs"
and/or aliases so we can get things rolling)

* support `iinfo` and `finfo` functions required by the standard:
https://data-apis.org/array-api/2021.12/API_specification/data_type_functions.html#objects-in-api

* needed to add a (hacky) implemention of `asarray()` to comply
with the array API standard per:
https://data-apis.org/array-api/2021.12/API_specification/creation_functions.html#objects-in-api
Note that the positional-only and keyword-only function signature
is also mandatory.

* the array API test suite detected multiple issue in our data type
system; these mostly seemed to stem from having both `DataType(Enum)`
and `DataTypeClass`, which is a model that is not consistent with
the mappings expected by the standard; it is also mandatory to support
more types than we currently do, so I've hacked around some of these
issues for now, but with boolean type we'll currently fail the array API tests

* we'll need to double check what to do for `0` dimensional arrays, but
they are used in the array API test suite, so I've added a temporary
hack around those until we support them "natively?"

* add a CI job that starts to test for array API compliance--it will
fail, but it should at least start running rather than erroring out
at the import stage, which is a step forward from `develop` branch

* the usual mypy ignores, at least for now..

### DIFF
--- a/pykokkos/interface/views.py
+++ b/pykokkos/interface/views.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 import ctypes
+import os
 import math
 from enum import Enum
 import sys
@@ -10,6 +11,7 @@ from typing import (
 
 import numpy as np
 
+import pykokkos as pk
 from pykokkos.bindings import kokkos
 import pykokkos.kokkos_manager as km
 
@@ -131,6 +133,8 @@ class ViewType:
         :returns: the length of the first dimension
         """
 
+        if len(self.shape) == 0:
+            return 0
         return self.shape[0]
 
     def __iter__(self) -> Iterator:
@@ -240,6 +244,8 @@ class View(ViewType):
         """
 
         self.shape: List[int] = shape
+        if self.shape == [0]:
+            self.shape = ()
         self.size = math.prod(shape)
         self.dtype: Optional[DataType] = self._get_type(dtype)
         if self.dtype is None:
@@ -259,9 +265,19 @@ class View(ViewType):
         self.layout: Layout = layout
         self.trait: Trait = trait
 
+        if self.dtype == pk.float:
+            self.dtype = DataType.float
+        elif self.dtype == pk.double:
+            self.dtype = DataType.double
+        elif self.dtype == pk.int32:
+            self.dtype = DataType.int32
+        elif self.dtype == pk.int64:
+            pass
         if trait is trait.Unmanaged:
             self.array = kokkos.unmanaged_array(array, dtype=self.dtype.value, space=self.space.value, layout=self.layout.value)
         else:
+            if len(shape) == 0:
+                shape = [1]
             self.array = kokkos.array("", shape, None, None, self.dtype.value, space.value, layout.value, trait.value)
         self.data = np.array(self.array, copy=False)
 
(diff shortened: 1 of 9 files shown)


pykokkos/interface/views.py
```diff
@@ -417,7 +449,16 @@ def from_numpy(array: np.ndarray, space: Optional[MemorySpace] = None, layout: O
     if layout is None:
         layout = Layout.LayoutDefault
 
-    return View(list(array.shape), dtype, space=space, trait=Trait.Unmanaged, array=array, layout=layout)
+    # TODO: pykokkos support for 0-D arrays?
+    # temporary/terrible hack here for array API testing..
```
tylerjereddy, line 453: So, the `pykokkos` tests proper are passing in this PR now. Of course, even the small subset of array API tests I've added in CI don't pass yet. There appear to be two main challenges:

1. Lack of native support for `bool` in Kokkos proper? I asked about this on the main Slack channel and there didn't seem to be any suggestions.
2. What we currently have exposed for 0-d array support. There were some comments on Slack which I'll just paste below:

> Kokkos has rank-0 View<T> which contains a single scalar value. It doesn’t have a unit-like View<void> or whatever

> yeah a rank-0 view is also accessible as you suggested by ()

```
View<int> a("A");
a() = 5;
```

> mdspan is the same (just with [] operator instead)

```
mdspan<int, dextents<0>> a(ptr_a);
a[] = 5;
```

tylerjereddy:
@NaderAlAwar Ok, the CI is passing now, so this is probably ready for review. I'll summarize what's going on here below:

- I ended up selecting to run only a small subset of the array API testsuite in CI to start--this is still progress in the sense that if you try to run the array API testsuite on `develop`, you won't even be able to import the tests let alone run/pass them because of the `dtype` system inconsistencies
- some of the changes still don't feel quite right, so those cases may also reflect situations where the current `pykokkos` testsuite isn't catching some stuff? see the selective `pass` here for example: https://github.com/kokkos/pykokkos/pull/57/files#diff-a471f687656d8dda27306cb2a3a4ec31fa9b182df23b5dbd72a563ac309ee83cR275
- we may want to open issues for some of the discussion points I made above re: `bool` and `0-D` array support; conversely, there are so many TODOs left on the array API conformance that opening issues for them all may be a bit much anyway for now
- I didn't have to modify any `pykokkos` tests proper, so that's probably a good sign
- there are probably a few extraneous imports in the diff, from a quick scan of my work

NaderAlAwar (commented):
The changes look good. I think eventually we need to clean up the data type issue where we have both the enum and the empty classes and address the bool and 0d array issues, but we can merge this now if you want.

PR #67, opened 2022-08-26
ENH, TST: more 0-D handling

tylerjereddy:
* we can now pass an additional testing module
in the array API suite, `test_utils.py`, which
has also been added to the CI workflow now

* `DataTypeClass` subclasses have been given
a `np_equiv` attribute to facilitate interopartion
with NumPy types; admittedly, this is probably
just a temporary adjustment as the type system
matures

* `View` instances now have slightly more robust
`0-D` array handling because of two adjustments:
  - lean a bit more heavily on NumPy in these cases
  - start using a custom `__eq__` for scalar comparisons (which
apparently also requires a custom `__hash__` for other
parts of `pykokkos`)

* my primary metric that this is actually "progress" instead
of just leaning more heavily on NumPy is that we can pass
more array API tests now; ultimately, this may just
result in deferring proper implementations for some things,
though being able to satisfy the array API with some parts
falling back to NumPy probably isn't a bad thing anyway, at
least at first

* no changes to the `pykokkos` test suite were needed


tylerjereddy:
Fixed up/updated a few more things and rebased while I was at it.

PR #73, opened 2022-08-30
ENH: pk.ones() to API standard

### LINKED ISSUE #70: ENH: missing pk.ones()
As revealed by `pytest array_api_tests/test_creation_functions.py::test_ones`. For now, a natural place to implement this function is probably `pykokkos/lib/create.py` next to `pk.zeros()`.

tylerjereddy:
Fixes #70

* implement `pk.ones()` such that
`array_api_tests/test_creation_functions.py::test_ones` passes

* the Python array API standard `test_ones()` makes use
of `pk.equal()`, which wasn't implemented, so an early-stage
implementation is added here (which should help later on as well,
since the need to occasionally do `view1 == view2` should be
obvious)

* this is no doubt a bit of a "hack," especially in terms
of full-featured broadcasting for `pk.equal`, which is currently only
supported for "scalar" broadcasting in my work here

* I was a bit surprised that I only needed to add limited 1D
support for `pk.equal()`, but I imagine the code there will grow
substantially to support more scenarios/remove hacks in the future

* note that this branch also borrows some of the typing
changes from gh-67 (and improves on them slightly I think)

* special shims are present for some `0-D` type scenarios,
and for now I've doubled down on the adoption of an unsigned
integer type as a proxy for a `bool` array--I suggest we at least
use a shorter width type for this in the future though, once
the pykokkos-base bindings are merged

* although I'm leaning on the Python array API standard tests here,
I did check that i.e., this works:

```python
import pykokkos as pk

def main():
    a = pk.ones((3, 3))
    print(a)

if __name__ == "__main__":
    main()
```

```
[[1. 1. 1.]
 [1. 1. 1.]
 [1. 1. 1.]]
```

* note, however, that the hook-in of `pk.equal()` to `__equal__`
wasn't the focus of this work (since the array API tests simply
call `pk.equal()` for the isolated test I was working on), so
for now:

```python
import pykokkos as pk

def main():
    a = pk.ones((3,))
    b = pk.ones((3,))
    result = a == b
    print(result)
    result2 = pk.equal(a, b)
    print(result2)

if __name__ == "__main__":
    main()
```

Produces:

```
False
[1 1 1]
```

* what we really want is `True` and `[True True True]`,
but the latter isn't too bad at least--the `1` values are
a proxy for truth in the `pk.uint16` type we're using
in place of `bool` for now; I suggest we delay the hook-in
to `__equal__` for now...

### DIFF
--- a/pykokkos/interface/views.py
+++ b/pykokkos/interface/views.py
@@ -181,7 +181,7 @@ class ViewType:
 class View(ViewType):
     def __init__(
         self,
-        shape: Tuple[int],
+        shape: Union[List[int], Tuple[int]],
         dtype: Union[DataTypeClass, type] = real,
         space: MemorySpace = MemorySpace.MemorySpaceDefault,
         layout: Layout = Layout.LayoutDefault,
@@ -191,7 +191,7 @@ class View(ViewType):
         """
         View constructor.
 
-        :param shape: the shape of the view as a tuple of integers
+        :param shape: the shape of the view as a list or tuple of integers
         :param dtype: the data type of the view, either a pykokkos DataType or "int" or "float".
         :param space: the memory space of the view. Will be set to the execution space of the view by default.
         :param layout: the layout of the view in memory.
@@ -256,7 +256,7 @@ class View(ViewType):
         """
         Initialize the view
 
-        :param shape: the shape of the view as a tuple of integers
+        :param shape: the shape of the view as a list or tuple of integers
         :param dtype: the data type of the view, either a pykokkos DataType or "int" or "float".
         :param space: the memory space of the view. Will be set to the execution space of the view by default.
         :param layout: the layout of the view in memory.
@@ -265,8 +265,6 @@ class View(ViewType):
         """
 
         self.shape: Tuple[int] = tuple(shape)
-        if self.shape == (0,):
-            self.shape = ()
         self.size: int = math.prod(shape)
         self.dtype: Optional[DataType] = self._get_type(dtype)
         if self.dtype is None:
@@ -286,9 +284,6 @@ class View(ViewType):
         self.layout: Layout = layout
         self.trait: Trait = trait
 
-        # TODO: if ufuncs stop inspecting
-        # type "strings," we should be able to
-        # purge these mappings
         if self.dtype == pk.float:
             self.dtype = DataType.float
         elif self.dtype == pk.double:
--- a/pykokkos/interface/views.py
+++ b/pykokkos/interface/views.py
@@ -22,7 +22,7 @@ from .data_types import (
     int16, int32, int64,
     uint8,
     uint16, uint32, uint64,
-    double
+    double, float64,
 )
 from .layout import get_default_layout, Layout
 from .memory_space import get_default_memory_space, MemorySpace
@@ -138,8 +138,14 @@ class ViewType:
         :returns: the length of the first dimension
         """
 
+        # NOTE: careful with 0-D treatments and __bool__
+        # related handling; you can have shape () and
+        # still be True for example...
         if len(self.shape) == 0:
-            return 0
+            if self.data != 0:
+                return 1
+            else:
+                return 0
         return self.shape[0]
 
     def __iter__(self) -> Iterator:
@@ -149,7 +155,11 @@ class ViewType:
         :returns: an iterator over the data
         """
 
-        return (n for n in self.data)
+        if self.data.ndim > 0:
+            return (n for n in self.data)
+        else:
+            # 0-D case returns empty generator
+            return zip()
 
     def __str__(self) -> str:
         """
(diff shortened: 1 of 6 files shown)


tylerjereddy:
I rebased and fixed a bunch of merge conflicts, since other PRs related to the `shape` object and so on were merged recently.

Locally, both the subset of the array API suite we test against and the regular test suite seem happy again, but we'll see if CI agrees...

tylerjereddy:
@NaderAlAwar I rebased and resolved merge conflicts. The CI is passing but I think there may be two things to consider/review here:

- the `dtype` class changes
- I think a bit more `pk.uint16` -> `pk.uint8` modernization might be in order for the bool proxies, given recent changes in `develop`

pykokkos/lib/create.py
```diff
@@ -2,3 +2,15 @@
...
+
+def ones(shape, *, dtype=None, device=None):
+    if dtype is None:
+        # NumPy also defaults to a double for ones()
+        dtype = pk.float64
+    view: pk.View = pk.View([*shape], dtype=dtype)
+    view[:] = 1
+    if shape == (0,):
+        view.shape = (0,)
+    view.shape = tuple(view.shape)
```
NaderAlAwar, line 15: I'm a bit confused by what you're trying to do with `view.shape` here. In `_init_view()` we currently do the following:
```Python
        self.shape: Tuple[int] = tuple(shape)
        if self.shape == (0,):
            self.shape = ()
```
Did you make these changes before we changed the shape to a tuple? If so, are they still needed? I'm also not sure why we're forcing `view.shape` to be `(0,)` in the if block

pykokkos/lib/ufuncs.py
```diff
@@ -1429,3 +1431,172 @@ def isinf(view):
...
+        view2_idx = tid
+    if view1[tid] == view2[view2_idx]:
+        view_result[tid] = 1
+    else:
+        view_result[tid] = 0
+
+def equal(view1, view2):
+    # TODO: write even more dispatching for cases where view1 and view2
+    # have different, but comparable, types (like float32 vs. float64?)
+    # this may "explode" without templating
```
NaderAlAwar, line 1522: I think this is fine for now. I'm still undecided on how to properly fix this, but I'm leaning towards getting rid of the type annotations.

pykokkos/lib/ufuncs.py
```diff
@@ -1429,3 +1431,172 @@ def isinf(view):
...
+    if view1.shape != view2.shape:
+        if not view1.size <= 1 and not view2.size <= 1:
+            # TODO: supporting __eq__ over broadcasted shapes beyond
+            # scalar (i.e., matching number of columns)
+            raise ValueError("view1 and view2 have incompatible shapes")
+
+    # TODO: something more appropriate than uint16 as a proxy
+    # for the bool type? (a shorter integer like uint8
+    # at least?)
+    view_result = pk.View([*view1.shape], dtype=pk.uint16)
```
NaderAlAwar, line 1536: We have `uint8` now right? Can we just use that?

pykokkos/lib/ufuncs.py
```diff
@@ -1429,3 +1431,172 @@ def isinf(view):
...
+    except AttributeError:
+        pass
+    try:
+        if isinstance(view1.array, np.ndarray):
+            if view1.shape == () or view1.shape == (0,):
+                view1r = pk.View([1], dtype=view1.dtype)
+                view1r[:] = view1.array
+                view1 = view1r
+    except AttributeError:
+        pass
```
NaderAlAwar, line 1559: Is this for the specific case where the views are created from a numpy array? Why do we need to allocate a new view and copy it over?

NaderAlAwar (commented):
Looks good overall, just need clarification on some of the comments I left

pykokkos/interface/views.py
```diff
@@ -171,7 +181,7 @@ def __deepcopy__(self, memo):
 class View(ViewType):
     def __init__(
         self,
-        shape: List[int],
+        shape: Tuple[int],
```
NaderAlAwar, line 184: Should be a Union of List and Tuple

pykokkos/interface/views.py
```diff
@@ -181,7 +191,7 @@ def __init__(
         """
         View constructor.
 
-        :param shape: the shape of the view as a list of integers
+        :param shape: the shape of the view as a tuple of integers
```
NaderAlAwar, line 194: Update to list or tuple

pykokkos/interface/views.py
tylerjereddy: done, I think

pykokkos/interface/views.py
tylerjereddy: done, I think

pykokkos/lib/ufuncs.py
tylerjereddy: I deleted the pertinent NumPy array specific code here--looks like we don't have tests that require that at this time

NaderAlAwar (approved):

PR #79, opened 2022-09-07
Multiple arithmetic, logical, and trigonometric ufuncs

ShahzadUmair:
The ufuncs this PR covers include:

- true_divide
- logaddexp2
- floor_divide
- sin
- cos
- tan
- logical_and
- logical_or
- logical_xor
- logical_not
- fmax
- fmin
- exp
- exp2

As of now these ufuncs only supports 1D (float and double) Views as input .

### DIFF
--- a/pykokkos/lib/ufuncs.py
+++ b/pykokkos/lib/ufuncs.py
@@ -948,7 +948,7 @@ def sin_impl_1d_float(tid: int, view: pk.View1D[pk.float], out: pk.View1D[pk.flo
 
 def sin(view):
     """
-    Element-wise trignometric sine of the view
+    Element-wise trigonometric sine of the view
 
     Parameters
     ----------
@@ -982,7 +982,7 @@ def cos_impl_1d_float(tid: int, view: pk.View1D[pk.float], out: pk.View1D[pk.flo
 
 def cos(view):
     """
-    Element-wise trignometric cosine of the view
+    Element-wise trigonometric cosine of the view
 
     Parameters
     ----------
@@ -1217,12 +1217,12 @@ def logical_not(view):
 
 
 @pk.workunit
-def fmax_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.uint16]):
+def fmax_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.double]):
     out[tid] = fmax(viewA[tid], viewB[tid])
 
 
 @pk.workunit
-def fmax_impl_1d_float(tid: int, viewA: pk.View1D[pk.float], viewB: pk.View1D[pk.float], out: pk.View1D[pk.uint16]):
+def fmax_impl_1d_float(tid: int, viewA: pk.View1D[pk.float], viewB: pk.View1D[pk.float], out: pk.View1D[pk.float]):
     out[tid] = fmax(viewA[tid], viewB[tid])
 
 
@@ -1239,12 +1239,12 @@ def fmax(viewA, viewB):
 
     Returns
     -------
-    out : pykokkos view (uint16)
+    out : pykokkos view
            Output view.
 
     """
-    out = pk.View([viewA.shape[0]], pk.uint16)
     if str(viewA.dtype) == "DataType.double" and str(viewB.dtype) == "DataType.double":
+        out = pk.View([viewA.shape[0]], pk.double)
         pk.parallel_for(
             viewA.shape[0],
             fmax_impl_1d_double,
@@ -1253,6 +1253,7 @@ def fmax(viewA, viewB):
             out=out)
 
     elif str(viewA.dtype) == "DataType.float" and str(viewB.dtype) == "DataType.float":
+        out = pk.View([viewA.shape[0]], pk.float)
         pk.parallel_for(
             viewA.shape[0],
             fmax_impl_1d_float,
@@ -1265,12 +1266,12 @@ def fmax(viewA, viewB):
 
 
 @pk.workunit
-def fmin_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.uint16]):
+def fmin_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.double]):
     out[tid] = fmin(viewA[tid], viewB[tid])
 
 
 @pk.workunit
-def fmin_impl_1d_float(tid: int, viewA: pk.View1D[pk.float], viewB: pk.View1D[pk.float], out: pk.View1D[pk.uint16]):
+def fmin_impl_1d_float(tid: int, viewA: pk.View1D[pk.float], viewB: pk.View1D[pk.float], out: pk.View1D[pk.float]):
     out[tid] = fmin(viewA[tid], viewB[tid])
 
 
@@ -1287,12 +1288,12 @@ def fmin(viewA, viewB):
 
     Returns
     -------
-    out : pykokkos view (uint16)
+    out : pykokkos view
            Output view.
 
     """
-    out = pk.View([viewA.shape[0]], pk.uint16)
     if str(viewA.dtype) == "DataType.double" and str(viewB.dtype) == "DataType.double":
+        out = pk.View([viewA.shape[0]], pk.double)
         pk.parallel_for(
             viewA.shape[0],
             fmin_impl_1d_double,
@@ -1301,6 +1302,7 @@ def fmin(viewA, viewB):
             out=out)
 
     elif str(viewA.dtype) == "DataType.float" and str(viewB.dtype) == "DataType.float":
+        out = pk.View([viewA.shape[0]], pk.float)
         pk.parallel_for(
             viewA.shape[0],
             fmin_impl_1d_float,
--- a/pykokkos/lib/ufuncs.py
+++ b/pykokkos/lib/ufuncs.py
@@ -409,9 +409,11 @@ def subtract(viewA, viewB):
         raise RuntimeError("Incompatible Types")
     return out
 
+
 @pk.workunit
 def matmul_impl_1d_double(tid: int, acc: pk.Acc[pk.double], viewA: pk.View1D[pk.double], viewB: pk.View2D[pk.double]):
     acc += viewA[tid] * viewB[0][tid]
+
 
 @pk.workunit
 def matmul_impl_1d_float(tid: int, acc: pk.Acc[pk.float], viewA: pk.View1D[pk.float], viewB: pk.View2D[pk.float]):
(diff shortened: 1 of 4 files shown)


pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+    """
+
+    return divide(viewA, viewB)
+
+
+@pk.workunit
+def logaddexp2_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.double],):
+    out[tid] = log2(pow(2, viewA[tid]) + pow(2, viewB[tid]))
+
+@pk.workunit
```
gliga, line 828: should there be an extra newline here

pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+    else:
+        raise RuntimeError("Incompatible Types")
+    return out
+
+
+@pk.workunit
+def floor_divide_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.double]):
+    out[tid] = viewA[tid] // viewB[tid]
+
+@pk.workunit
```
gliga, line 877: same question

pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+def sin_impl_1d_double(tid: int, view: pk.View1D[pk.double], out: pk.View1D[pk.double]):
+    out[tid] = sin(view[tid])
+
+@pk.workunit
+def sin_impl_1d_float(tid: int, view: pk.View1D[pk.float], out: pk.View1D[pk.float]):
+    out[tid] = sin(view[tid])
+
+def sin(view):
```
gliga, line 930: same question

pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+    else:
+        raise RuntimeError("Incompatible Types")
+    return out
+
+
+@pk.workunit
+def sin_impl_1d_double(tid: int, view: pk.View1D[pk.double], out: pk.View1D[pk.double]):
+    out[tid] = sin(view[tid])
+
+@pk.workunit
```
gliga, line 926: same

pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+    view : pykokkos view
+           Input view.
+
+    Returns
+    -------
+    out : pykokkos view
+           Output view.
+
+    """
+    if str(view.dtype) == "DataType.double":
```
gliga, line 945: why do we need to convert to string?

pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+            Input view.
+
+    Returns
+    -------
+    out : pykokkos view (uint16)
+           Output view.
+
+    """
+    out = pk.View([viewA.shape[0]], pk.uint16)
+    if str(viewA.dtype) == "DataType.double" and str(viewB.dtype) == "DataType.double":
```
gliga, line 1088: instead of doing all these comparison with strings, we should introduce some global functions like
`type_eq(viewA.dtype, DataType.double)`, then inside the function convert to anything if necessary
@NaderAlAwar

pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+           Output view.
+
+    """
+    if str(view.dtype) == "DataType.double":
+        out = pk.View([view.shape[0]], pk.double)
+        pk.parallel_for(view.shape[0], tan_impl_1d_double, view=view, out=out)
+    elif str(view.dtype) == "DataType.float":
+        out = pk.View([view.shape[0]], pk.float)
+        pk.parallel_for(view.shape[0], tan_impl_1d_float, view=view, out=out)
+    return out
```
gliga, line 1013: why is there no else and exception in this case?

pykokkos/lib/ufuncs.py
tylerjereddy: Good question--we've been doing this because we don't really have proper type object/type hierarchy implemented yet--you can see that we just have some "skeletons" in https://github.com/kokkos/pykokkos/blob/develop/pykokkos/interface/data_types.py. 

We probably eventually want something like [`np.can_cast`](https://numpy.org/doc/stable/reference/generated/numpy.can_cast.html) for casting-compatible hierarchy assessment.

I think Nader and I both liked the type promotion lattice that Google uses for Jax: https://jax.readthedocs.io/en/latest/type_promotion.html?highlight=types (and they are also concerned with the array API conformance I think).

pykokkos/lib/ufuncs.py
NaderAlAwar: You're right, it should be added here

pykokkos/lib/ufuncs.py
NaderAlAwar: Agreed, I opened #80 to track this. I think we can resolve this for this PR.

tests/test_ufuncs.py
```diff
@@ -322,7 +328,7 @@ def test_1d_exposed_ufuncs_vs_numpy(pk_ufunc,
     actual = pk_ufunc(view=view)
     # log10 single-precision needs relaxed tol
     # for now
-    if numpy_ufunc == np.log10 and numpy_dtype == np.float32:
+    if numpy_ufunc in [np.log10, np.cos] and numpy_dtype == np.float32:
```
NaderAlAwar, line 331: I think you should use a set here, so `in {np.log10, np.cos}`.

NaderAlAwar (changes requested):
Looks good, some minor changes needed.

pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+        raise RuntimeError("Incompatible Types")
+    return out
+
+@pk.workunit
+def fmin_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.uint16]):
+    out[tid] = fmin(viewA[tid], viewB[tid])
+
+@pk.workunit
+def fmin_impl_1d_float(tid: int, viewA: pk.View1D[pk.float], viewB: pk.View1D[pk.float], out: pk.View1D[pk.uint16]):
+    out[tid] = fmin(viewA[tid], viewB[tid])
```
tylerjereddy, line 1235: Note that the CI is failing on this line:

```
Error on line 1235 : Function fmin not supported for translation
DEBUG AST:
Name(id='fmin', ctx=Load())
    out[tid] = fmin(viewA[tid], viewB[tid])
               ^^^^
Translation of <ast.FunctionDef object at 0x7f495e0dd3f0> failed
```

pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+        out = pk.View([view.shape[0]], pk.float)
+        pk.parallel_for(view.shape[0], logical_not_impl_1d_float, view=view, out=out)
+    return out
+
+@pk.workunit
+def fmax_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.uint16]):
+    out[tid] = fmax(viewA[tid], viewB[tid])
+
+@pk.workunit
+def fmax_impl_1d_float(tid: int, viewA: pk.View1D[pk.float], viewB: pk.View1D[pk.float], out: pk.View1D[pk.uint16]):
```
tylerjereddy, line 1189: Wrong type for `out` here--see my comment on `fmin()`

pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+def sin_impl_1d_double(tid: int, view: pk.View1D[pk.double], out: pk.View1D[pk.double]):
+    out[tid] = sin(view[tid])
+
+@pk.workunit
+def sin_impl_1d_float(tid: int, view: pk.View1D[pk.float], out: pk.View1D[pk.float]):
+    out[tid] = sin(view[tid])
+
+def sin(view):
+    """
+    Element-wise trignometric sine of the view
```
tylerjereddy, line 932: here and elsewhere: trignometric typo

pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+        out = pk.View([view.shape[0]], pk.float)
+        pk.parallel_for(view.shape[0], tan_impl_1d_float, view=view, out=out)
+    return out
+
+@pk.workunit
+def logical_and_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.uint16]):
+    out[tid] = viewA[tid] and viewB[tid]
+
+@pk.workunit
+def logical_and_impl_1d_float(tid: int, viewA: pk.View1D[pk.float], viewB: pk.View1D[pk.float], out: pk.View1D[pk.uint16]):
```
tylerjereddy, line 1020: `pk.uint8` was merged in gh-77, so we may as well start using that as a smaller proxy for `bool`?

pykokkos/lib/ufuncs.py
```diff
@@ -798,4 +798,538 @@ def logaddexp(viewA, viewB):
...
+    else:
+        raise RuntimeError("Incompatible Types")
+    return out
+
+@pk.workunit
+def logical_not_impl_1d_double(tid: int, view: pk.View1D[pk.double], out: pk.View1D[pk.double]):
+    out[tid] = not view[tid]
+
+@pk.workunit
+def logical_not_impl_1d_float(tid: int, view: pk.View1D[pk.float], out: pk.View1D[pk.float]):
```
tylerjereddy, line 1158: Shouldn't `logical_not` have a return (`out`) type that is `bool`, which for us means `pk.uint8`?

tylerjereddy: It depends how well we want to do it I guess. How many bits are used for the C++ bool? The `np.bool_` uses `8` bits I think, which may surprise some. Conversely, the type/`struct` may need to have a bit more metadata than just `1` or `0`, and I don't think Kokkos at C++ level uses `bool` directly in parallel loops?

Maybe we basically use `1` and `0` under the hood with a few "decorations" for printing them as `True` and `False` at the end of the day.

gliga: Let's just have it as an alias to `int8` and then we can later change what is in background

gliga (approved):
Looks good to me

NaderAlAwar (approved):

PR #82, opened 2022-09-12
ENH: add API constants

### LINKED ISSUE #68: TST, ENH: some missing constants
`pytest array_api_tests/test_constants.py` reveals that we need to add constants for:

- [ ] `pk.nan`
- [ ] `pk.inf`
- [ ] `pk.pi`
- [ ] `pk.e`

tylerjereddy:
* Fixes #68 by adding the `pk.e`, `pk.pi`, `pk.inf`, and `pk.nan` constants needed to pass `array_api_tests/test_constants.py` (we'll see if the CI agrees with me...)

* the implementations are a bit crude, but do seem to pass tests locally at least

* note that some of the infrastructure here is borrowed from other unmerged PRs

### DIFF
--- a/pykokkos/interface/views.py
+++ b/pykokkos/interface/views.py
@@ -532,7 +532,7 @@ def asarray(obj, /, *, dtype=None, device=None, copy=None):
     # TODO: proper implementation/design
     # for now, let's cheat and use NumPy asarray() followed
     # by pykokkos from_numpy()
-    if obj in [pk.e, pk.pi, pk.inf, pk.nan]:
+    if obj in {pk.e, pk.pi, pk.inf, pk.nan}:
         view = pk.View([1], dtype=pk.float32)
         view[:] = obj
         return view
--- a/pykokkos/interface/views.py
+++ b/pykokkos/interface/views.py
@@ -532,6 +532,10 @@ def asarray(obj, /, *, dtype=None, device=None, copy=None):
     # TODO: proper implementation/design
     # for now, let's cheat and use NumPy asarray() followed
     # by pykokkos from_numpy()
+    if obj in [pk.e, pk.pi, pk.inf, pk.nan]:
+        view = pk.View([1], dtype=pk.float32)
+        view[:] = obj
+        return view
     if "bool" in str(dtype):
         dtype = np.bool_
     arr = np.asarray(obj, dtype=dtype)
--- a/pykokkos/interface/views.py
+++ b/pykokkos/interface/views.py
@@ -533,7 +533,9 @@ def asarray(obj, /, *, dtype=None, device=None, copy=None):
     # for now, let's cheat and use NumPy asarray() followed
     # by pykokkos from_numpy()
     if obj in {pk.e, pk.pi, pk.inf, pk.nan}:
-        view = pk.View([1], dtype=pk.float32)
+        if dtype is None:
+            dtype = pk.float64
+        view = pk.View([1], dtype=dtype)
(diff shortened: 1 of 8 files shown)


pykokkos/lib/constants.py
```diff
@@ -0,0 +1,6 @@
+import math
+
+e = math.e
+pi = math.pi
+inf = math.inf
+nan = math.nan
```
tylerjereddy, line 6: this feels a bit lazy, although locally the array API suite didn't seem to complain

tylerjereddy:
I'll have to look into the array API failure, and what differs vs. my local scenario.

tylerjereddy:
To summarize the additional changes that were needed to get CI passing here:

1. `*.sh` files were not installed by the `pykokkos` `setup.py`, which was a real bug related to https://github.com/kokkos/pykokkos/issues/3#issuecomment-1038457361 (in short, we don't really test *installed* `pykokkos` that well yet, because we lean on `python runtests.py` from the source dir..); more issues will likely show up in the future as we flush through more and more of the array API suite
2. Pinning Python version to avoid a GitHub actions issue in gh-83.

pykokkos/interface/views.py
```diff
@@ -532,6 +532,10 @@ def asarray(obj, /, *, dtype=None, device=None, copy=None):
     # TODO: proper implementation/design
     # for now, let's cheat and use NumPy asarray() followed
     # by pykokkos from_numpy()
+    if obj in [pk.e, pk.pi, pk.inf, pk.nan]:
```
NaderAlAwar, line 535: Minor comment, but isn't it better to use a `set` here? So `{pk.e, pk.pi, pk.inf, pk.nan}`

NaderAlAwar (changes requested):
Looks good, I'll merge once you replace the list with the set (if you think that's necessary)

tylerjereddy:
I changed the list to a set op--it is a bit faster with the hashed constant time lookup yeah.

A bigger problem is that this implementation of `asarray()` forces you to use `float32` as the type with the new constants, even though `float64` should be perfectly valid for i.e., `pk.nan` as well. It also ignores the user specified `dtype` when handling those constants.

So, it is really just an attempt to pass more array API tests rather than anything close to a final implementation.

pykokkos/interface/views.py
```diff
@@ -532,6 +532,10 @@ def asarray(obj, /, *, dtype=None, device=None, copy=None):
     # TODO: proper implementation/design
     # for now, let's cheat and use NumPy asarray() followed
     # by pykokkos from_numpy()
+    if obj in {pk.e, pk.pi, pk.inf, pk.nan}:
+        view = pk.View([1], dtype=pk.float32)
```
NaderAlAwar, line 536: Regarding the data type issue you mentioned, couldn't you just use the `dtype` argument if it's set to `float64`, and then have it default to `float32` otherwise?

NaderAlAwar:
Thanks Tyler. I added a comment regarding the data type issue, let me know what you think

tylerjereddy:
We could do that, though we probably also need error handling if you request an integer type for the floating consts.

NaderAlAwar:
How do the other array libraries handle this? Do they default to `float32`? Do they allow `float64`?

tylerjereddy:
@NaderAlAwar I added a test and changes to match NumPy for those constants with `asarray()`

NaderAlAwar (approved):

Issue #62, opened 2022-08-25
ENH: ufunc implementation tracking list (up to 3D)


tylerjereddy:
Tracking issue for `ufunc` implementations (see: https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists ). 

- [ ] [numpy.add](https://numpy.org/doc/stable/reference/generated/numpy.add.html#numpy.add)
  - [ ] 1-D
  - [ ] 2-D
  - [ ] 3-D
- [ ] [numpy.subtract](https://numpy.org/doc/stable/reference/generated/numpy.subtract.html#numpy.subtract)
  - [ ] 1-D
  - [ ] 2-D
  - [ ] 3-D
- [ ] [numpy.multiply](https://numpy.org/doc/stable/reference/generated/numpy.multiply.html#numpy.multiply)
  - [ ] 1-D
  - [ ] 2-D
  - [ ] 3-D
- [ ] [numpy.matmul](https://numpy.org/doc/stable/reference/generated/numpy.matmul.html#numpy.matmul)
  - [ ] 1-D
  - [ ] 2-D
  - [ ] 3-D
- [ ] [numpy.divide](https://numpy.org/doc/stable/reference/generated/numpy.divide.html#numpy.divide)
  - [ ] 1-D
  - [ ] 2-D
  - [ ] 3-D
- [ ] [numpy.logaddexp](https://numpy.org/doc/stable/reference/generated/numpy.logaddexp.html#numpy.logaddexp)
  - [ ] 1-D
  - [ ] 2-D
  - [ ] 3-D
- [ ] [numpy.logaddexp2](https://numpy.org/doc/stable/reference/generated/numpy.logaddexp2.html#numpy.logaddexp2)
  - [ ] 1-D
  - [ ] 2-D
  - [ ] 3-D
- [ ] [numpy.true_divide](https://numpy.org/doc/stable/reference/generated/numpy.true_divide.html#numpy.true_divide)
  - [ ] 1-D
  - [ ] 2-D
  - [ ] 3-D
- [ ] [numpy.floor_divide](https://numpy.org/doc/stable/reference/generated/numpy.floor_divide.html#numpy.floor_divide)
  - [ ] 1-D
  - [ ] 2-D
  - [ ] 3-D
- [ ] [numpy.negative](https://numpy.org/doc/stable/reference/generated/numpy.negative.html#numpy.negative)
  - [ ] 1-D
  - [ ] 2-D
  - [ ] 3-D
[...]

PR #84, opened 2022-09-16
Gaussian Naive Bayes example with pykokkos

ShahzadUmair:
This pr includes a basic implementation of Gaussian Naive Bayes with pykokkos. (Still working on polishing a few things)

**A few new ufuncs:**

**sum**: uses np.sum to calculate sum along the axis if axis is specified, otherwise uses parallel reduce to sum a 1D view.

**argmax**: uses np.argmax to find the indices of max values along an axis. This would rely on "max" fn in pykokkos for which we are still working on finding an optimal implementation

**unique**: using np.unique to find the sorted unique elements of a view. Actual implementation would require an optimal parallel sorting algorithm and filtering

**var**: using np.var to find variance along an axis in a 2d array.

**in1D**: Tests if an element of the view is present in the other view. Returns view of 1st view's shape.

**mean**: uses `sum` and `divide_num` functions to find the mean along given axis.

**divide_num**: divides a view with a scalar. Currently only supports upto 2d pk.double views.

**mul_num**: multiplies a view with a scalar. Currently only supports 1D pk.double views.

**add_num**: adds a scalar to every element of a view. Currently only supports upto 2d pk.double views

**transpose**: Serial transpose (uses the built-in zip function) of 1D views.

**power_num**: raises every element of the view to a scalar value. Currently only supports 1D pk.double views

**index**: replaces the elements(e) of a view with the values of another view (w)such that the final result is w[e]

**findMax**: return the max element of the view (curently uses the pythons builtin function)

**searchsorted**: linear search

The functions using the numpy implementation (except `var`) require implementation of parallel algorithms of their own. 

In addition, we are hard coding quite a bit by adding a lot of new code for making views work with different shaped views and scalars. Probably introducing broadcasting will help fix this issues.


pykokkos/lib/util.py
```diff
@@ -8,12 +8,14 @@
 # https://data-apis.org/array-api/2021.12/API_specification/utility_functions.html
 
 def all(x, /, *, axis=None, keepdims=False):
+    return np.all(x)
```
ShahzadUmair, line 11: `pk.all([1, 1, 1])`  where [1, 1, 1] is pk.View([3], pk.int8)

returns `False` since `[1,1,1] == True` is False. Added a temporary np.all here. Will fix it soon

pykokkos/__init__.py
```diff
@@ -45,8 +45,23 @@
...
                                  exp,
                                  exp2,
-                                isinf,
-                                isnan)
+                                 argmax,
+                                 unique,
+                                 var,
+                                 in1d,
+                                 mean,
+                                 sum,
```
tylerjereddy, line 53: Just a drive-by comment for now, but I'm pretty sure that `sum()` is not a `ufunc`--it shouldn't be listed in the NumPy docs under available ufuncs: https://numpy.org/doc/stable/reference/ufuncs.html#available-ufuncs

nor should it be listed as a ufunc in our checklist at gh-62. I think there's a distinction into a separate category for the pure reduction operations, so we may want to reflect that in our library/module organization.

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
     return out
 
 
+@pk.workunit
+def sum_impl_1d_double(tid: int, acc: pk.Acc[pk.double], viewA: pk.View1D[pk.double]):
+    acc += viewA[tid]
+
+
+def sum(viewA, axis=None):
+    if(axis is not None):
```
tylerjereddy, line 1398: I'd probably move `sum()` and other non-ufuncs elsewhere, and may lose the brackets here too--`if axis is not None:` would be a bit more idiomatic I think.

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+    rows = len(arr)
+    cols = 1
+
+    if (rows > 0 and isinstance(arr[0], Sequence)):
+        cols = len(arr[0])
+    
+    return [rows, cols] if (cols > 1) else [rows]
+
+
+def array(view):
```
tylerjereddy, line 1592: this is really strange, we already have `asarray()` and `from_numpy()`, which live elsewhere, so I can't imagine we'd want to have `array()`, and many of the other non-ufunc ops living in `ufuncs.py`. 

Importantly, what is `array()` giving you that `asarray()` is not? It looks like this implementation is less flexible as well--only supporting a single `dtype` vs. our `asarray()` implementation? We'd at least need to have docs explaining why `array()` needs to exist in this form, whithout leveraging the already-available `from_numpy()` and `asarray()` infrastructure.

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+            viewA.shape[0],
+            add_numm_impl_1d_double,
+            viewA=viewA,
+            viewB=viewB,
+            out=out)
+    else:
+        raise RuntimeError("Incompatible Types")
+    return out
+
+def _get_shape(arr):
```
tylerjereddy, line 1582: This looks like it is patching an oversight somewhere if you really need it? Can you summarize why our current size/shape infrastructure wasn't cutting if for you? I though we were doing pretty well in that regard apart from 0-d arrays maybe.

I'm also slightly confused by the design decision to return list objects when Nader just adjusted us to comply with the array API shape requirement of being tuple-like in gh-78.

tylerjereddy (commented):
Just added a few initial comments

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+
+    if str(viewA.dtype) == "DataType.double":
+        return pk.parallel_reduce(
+            viewA.shape[0],
+            sum_impl_1d_double,
+            viewA=viewA)
+
+
+def argmax(view, axis):
+    res = np.argmax(view, axis=axis)
```
NaderAlAwar, line 1431: This will only work if `view` is a numpy array, right? I think the assumption we're making with all these ufuncs is that all array inputs are Views. I understand that for now we need to use numpy, but even if `view` was of type View, you could still access it as a numpy array using the `data` field (see [here](https://github.com/kokkos/pykokkos/blob/a39282171630803067c8c2f13858e0a9beae2eb9/pykokkos/interface/views.py#L282)). So this would be `view.data` instead of just `view`.

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+    view = pk.View(res.shape, pk.int32)
+    view[:] = res
+
+    return view
+
+def unique(view):
+    res = np.unique(view)
```
NaderAlAwar, line 1438: Same comment as `argmax` (use `view.data`)

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+    view[:] = res
+
+    return view
+
+def var(view, axis):
+    res = np.var(view, axis=axis)
+    view = pk.View(res.shape, pk.double)
+    view[:] = res
+
+    return view
```
NaderAlAwar, line 1449: Why can't we implement our own variance kernel?

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+#     for i in range(viewA.extent(0)):
+#         out[tid] = out[tid] + (viewA[tid][i] / viewA.extent(0))
+
+
+# @pk.workunit
+# def mean1_____impl_1d_double(tid: int, viewA: pk.View2D[pk.double], out: pk.View1D[pk.double]):
+#     x = 0
+#     for i in range(viewA.extent(1)):
+#         x += (viewA[i][tid] / viewA.extent(1))
+#     out[tid] = x
```
NaderAlAwar: I assume this is a work in progress

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+
+
+
+def mean(view, axis):
+    viewA = pk.View(view.shape, pk.double)
+    viewA[:] = view
```
NaderAlAwar, line 1467: Slightly confused here, is `view` a numpy array, and are you converting from a numpy array to a view here? If so, just use `asarray()` (see [here](https://github.com/kokkos/pykokkos/blob/eb46e4b7f7ac9a0dc7c4b54fb15b631b9b244295/pykokkos/interface/views.py#L531))

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+
+    if str(viewA.dtype) == "DataType.double":
+        if (axis == 0):
+            sum_view = sum(viewA, axis=0)
+            return divide_num(sum_view, viewA.shape[0]) 
+        else:
+            sum_view = sum(viewA, axis=1)
+            return divide_num(sum_view, viewA.shape[1])
+    else:
+        raise RuntimeError("Incompatible Types")
```
NaderAlAwar, line 1480: @ShahzadUmair I think you'll fix this as discussed?

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+
+
+@pk.workunit
+def in1d_impl_1_d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.int8]):
```
NaderAlAwar, line 1484: Remove the underscore between 1 and d (so `in1d_impl_1d_double`)

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+    out[tid] = 0
+    for i in range(viewB.extent(0)):
+        if (viewB[i] == viewA[tid]):
+            out[tid] = 1
```
NaderAlAwar, line 1488: This might not be the fastest way to implement this but I think it's okay for now. I think you can add a `break` statement once you've found the element. Also, remove the parentheses in the if condition.

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+            viewA=viewA,
+            viewB=viewB,
+            out=out)
+    else:
+        raise RuntimeError("Incompatible Types")
+
+    return out
+
+@pk.workunit
+def divide_num2_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.double]):
```
NaderAlAwar, line 1505: A bit confused about the name, why is it `num2`?

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+
+
+@pk.workunit
+def divide_num_2d_impl_1d_double(tid: int, viewA: pk.View2D[pk.double], viewB: pk.View1D[pk.double], out: pk.View2D[pk.double]):
+    for i in range(viewA.extent(1)):
+        out[tid][i] = viewA[tid][i] / viewB[i]
+
+
+def divide_num(viewA, viewB):
+    if (not hasattr(viewB, '__iter__')):
```
NaderAlAwar, line 1516: We mentioned before that we will eventually do better type checking, but I think this can be improved for now by just using `if not isinstance(viewB, pk.View)`

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+        view_temp = pk.View([1], pk.double)
+        view_temp[0] = viewB
+        viewB = view_temp
+
+    if len(viewA.shape) == 2:
```
NaderAlAwar, line 1521: If you're checking the dimensionality of the view, just use `viewA.rank()` (see [here](https://github.com/kokkos/pykokkos/blob/eb46e4b7f7ac9a0dc7c4b54fb15b631b9b244295/pykokkos/interface/views.py#L70))

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+            viewB=viewB,
+            out=out)
+    elif str(viewA.dtype) == "DataType.double":
+        out = pk.View([viewA.shape[0]], pk.double)
+        pk.parallel_for(
+            viewA.shape[0],
+            divide_num2_impl_1d_double,
+            viewA=viewA,
+            viewB=view_temp,
+            out=out)
```
NaderAlAwar, line 1536: I'm not sure about these `if` `elif` conditions. Why do we need a special case for 2d arrays and then check for the datatype?

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+    out[tid] = viewA[tid] / viewB[0]
+
+
+@pk.workunit
+def divide_num_2d_impl_1d_double(tid: int, viewA: pk.View2D[pk.double], viewB: pk.View1D[pk.double], out: pk.View2D[pk.double]):
+    for i in range(viewA.extent(1)):
+        out[tid][i] = viewA[tid][i] / viewB[i]
+
+
+def divide_num(viewA, viewB):
```
NaderAlAwar, line 1515: Why do we have `divide_num` and not just divide?

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+        out = pk.View([viewc.shape[0]], pk.double)
+        pk.parallel_for(
+            viewA.shape[0],
+            mul0d_impl_1d_double,
+            viewA=viewc,
+            viewB=view_temp,
+            out=out)
+    else:
+        raise RuntimeError("Incompatible Types")
+    return out
```
NaderAlAwar, line 1563: Same comments as above

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+        out = pk.View([viewA.shape[0]], pk.double)
+        pk.parallel_for(
+            viewA.shape[0],
+            add_numm_impl_1d_double,
+            viewA=viewA,
+            viewB=viewB,
+            out=out)
+    else:
+        raise RuntimeError("Incompatible Types")
+    return out
```
NaderAlAwar, line 1599: Same comments as above

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+    view[:] = res
+
+    return view
+
+def transpose(view):
+    res = list(map(list, zip(*view)))
+    view = pk.View(_get_shape(res), pk.double)
+    view[:] = res
+
+    return view
```
NaderAlAwar, line 1623: Is it possible to implement this with a kernel?

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+
+@pk.workunit
+def power_num_impl_2d_double(tid: int, viewA: pk.View2D[pk.double], viewB: pk.View1D[pk.double], out: pk.View2D[pk.double]):
+    for i in range(viewA.extent(1)):
+        out[tid][i] = pow(viewA[tid][i], viewB[0])
+
+
+def power_num(viewA, viewB):
+    if (not hasattr(viewB, '__iter__')):
```
NaderAlAwar, line 1632: Same comment as above regarding checking the view type

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+            index_impl_1d_double,
+            viewA=viewA,
+            viewB=viewB,
+            out=out)
+    else:
+        raise RuntimeError("Incompatible Types")
+    return out
+
+
+def findMax(viewA):
```
NaderAlAwar, line 1668: where do we use this?

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1389,270 @@ def exp2(view):
...
+    return max(viewA)
+
+def searchsorted(view, ele):
```
NaderAlAwar, line 1671: where do we use this?

NaderAlAwar (changes requested):

pykokkos/lib/create.py
```diff
@@ -1,4 +1,7 @@
 import pykokkos as pk
 
-def zeros(shape, *, dtype=None, device=None):
-    return pk.View([*shape], dtype=dtype)
+def zeros(shape, *, dtype=pk.double, device=None):
+    if (hasattr(shape, '__iter__')):
+        return pk.View([*shape], dtype=dtype)
+    else:
+        return pk.View([shape], dtype=dtype)
```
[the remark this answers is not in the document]
tylerjereddy: The function signature should probably remain consistent with the array API standard requiring `dtype=None` as the default: https://data-apis.org/array-api/latest/API_specification/generated/signatures.creation_functions.zeros.html#signatures.creation_functions.zeros

though it likely doesn't matter too much. See the approach used for `pk.ones()` in gh-73. NumPy is a violator, though it predates the standard.

One motivating reason is in the design of newer array library like pytorch, where the default `dtype=None` can vary per the docs: https://pytorch.org/docs/stable/generated/torch.zeros.html

pykokkos/lib/ufuncs.py
ShahzadUmair: The type of view in this function is `pk.View2D[pk.double]`. The np.argmax function i think takes as input anything array-like so it is working out of the box thats why. Do you think i should still change it to view.data?

pykokkos/lib/ufuncs.py
ShahzadUmair: I have removed this code in the latest commit. This was extra, just making sure np arrays and pk views both work. Now we only have `pk.view` so it is working fine without this as well.

pykokkos/lib/ufuncs.py
ShahzadUmair: Done

pykokkos/lib/ufuncs.py
ShahzadUmair: Used in the gaussian bayes algo to find max for pk.view of variances

pykokkos/lib/ufuncs.py
ShahzadUmair: Gaussian Bayes use the search sorted function in `_partial_fit` function

examples/NaiveBayes/GaussianNB.py
```diff
@@ -0,0 +1,609 @@
+from abc import ABCMeta, abstractmethod
```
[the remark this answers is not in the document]
ShahzadUmair: Updated. Added at the top

pykokkos/lib/ufuncs.py
ShahzadUmair: Done

pykokkos/lib/ufuncs.py
ShahzadUmair: I wasnt aware about the pk_cpp folder back then. Forgot to revert this back after testing 😅

pykokkos/lib/ufuncs.py
ShahzadUmair: I have updated these, it was a buggy condition

pykokkos/lib/ufuncs.py
ShahzadUmair: I have moved the implementation to divide

pykokkos/lib/ufuncs.py
ShahzadUmair: Yes, done!

pykokkos/lib/ufuncs.py
ShahzadUmair: You are right, we dont require this. I made it during testing but forgot to remove.

pykokkos/lib/ufuncs.py
ShahzadUmair: Removed and moved all the lists to pk.view types. We dont need this function anymore.

pykokkos/interface/views.py
```diff
@@ -333,6 +333,10 @@ def _get_type(self, dtype: Union[DataType, type]) -> Optional[DataType]:
 
 
     def __eq__(self, other):
+        # Tempprary fix. Will change soon 
+        if other.dtype.type == np.float64:
+            return self.array == other
+
```
NaderAlAwar, line 339: Why/where is this needed?

pykokkos/interface/views.py
```diff
@@ -547,14 +554,21 @@ def from_cupy(array) -> ViewType:
...
 # asarray is required for comformance with the array API:
 # https://data-apis.org/array-api/2021.12/API_specification/creation_functions.html#objects-in-api
-
 def asarray(obj, /, *, dtype=None, device=None, copy=None):
     # TODO: proper implementation/design
     # for now, let's cheat and use NumPy asarray() followed
     # by pykokkos from_numpy()
-    if obj in {pk.e, pk.pi, pk.inf, pk.nan}:
+    if isHashable(obj) and obj in {pk.e, pk.pi, pk.inf, pk.nan}:
```
NaderAlAwar, line 571: Why do we need to check if it's hashable?

pykokkos/lib/ufuncs.py
```diff
@@ -1,4 +1,8 @@
+from typing import Sequence
+
 import pykokkos as pk
+import numpy as np
+import math
```
NaderAlAwar, line 5: Two things: move `import math` to the top, since it's a standard library import (numpy and pykokkos are 3rd party libraries), and import numpy before you import pykokkos (alphabetical order)

pykokkos/lib/ufuncs.py
```diff
@@ -84,12 +88,16 @@ def log(view):
...
+        return math.log(view)
+
     if str(view.dtype) == "DataType.double":
         pk.parallel_for(view.shape[0], log_impl_1d_double, view=view)
     elif str(view.dtype) == "DataType.float":
         pk.parallel_for(view.shape[0], log_impl_1d_float, view=view)
+    
     return view
-
+
```
NaderAlAwar, line 100: Remove the space character

pykokkos/lib/ufuncs.py
```diff
@@ -260,13 +268,18 @@ def sign(view):
 
 @pk.workunit
 def add_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.double], ):
-    out[tid] = viewA[tid] + viewB[tid]
+    out[tid] = viewA[tid] + viewB[tid % viewB.extent(0)]
```
NaderAlAwar, line 271: Is this needed if viewA and viewB have different sizes? Is this what numpy does?

pykokkos/lib/ufuncs.py
```diff
@@ -260,13 +268,18 @@ def sign(view):
...
 
 @pk.workunit
 def add_impl_1d_float(tid: int, viewA: pk.View1D[pk.float], viewB: pk.View1D[pk.float], out: pk.View1D[pk.float]):
     out[tid] = viewA[tid] + viewB[tid]
 
+@pk.workunit
+def add_num_2d_impl_1d_double(tid: int, viewA: pk.View2D[pk.double], viewB: pk.View1D[pk.double], out: pk.View2D[pk.double]):
+    for i in range(viewA.extent(1)):
+        out[tid][i] = viewA[tid][i] + viewB[i % viewB.extent(0)]
+
```
NaderAlAwar, line 282: Why does the workunit start with `add_num` instead of just `add`? Also same question for the modulo

pykokkos/lib/ufuncs.py
NaderAlAwar: If it's working right now it's fine, since we're going to be replacing this eventually

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1451,147 @@ def exp2(view):
...
+@pk.workunit
+def transpose_impl_2d(tid: int, view: pk.View2D[pk.double], out: pk.View2D[pk.double]):
+    for i in range(view.extent(1)):
+        out[i][tid] = view[tid][i]
+
+
+def transpose(view):
+    view_temp = pk.View(view.shape, dtype=pk.double)
+    view_temp[:] = view
+    view = view_temp
```
NaderAlAwar, line 1559: Is view_temp still needed?

pykokkos/lib/ufuncs.py
```diff
@@ -1381,6 +1451,147 @@ def exp2(view):
...
+
+    if view.rank() == 2:
+        if str(view.dtype) == "DataType.double":
+            out = pk.View(view.shape[::-1], pk.double)
+            pk.parallel_for(view.shape[0], transpose_impl_2d, view=view, out=out)
+            return out
+    
+    raise RuntimeError("Incompatible Types")
```
NaderAlAwar, line 1567: Replace the error message with something related to the view's shape (not being rank 2)

pykokkos/lib/ufuncs.py
NaderAlAwar: I think the original implementation called a `max()` method on the array itself, so we're changing the interface a bit here. I think this is fine for now, but can you rename this to `find_max` for now? Also add a TODO: comment at the point where this is used

NaderAlAwar (commented):
Good changes, I think you're almost there

pykokkos/interface/views.py
ShahzadUmair: at a few places the Naive Bayes code creates subviews like:

```python3
X[ y == y_i, :]  # X: pk.View2D[pk.double], y: pk.View1D[pk.double], y_i: Union(np.float64, pk.double)
```

I have updated this to make it more generic and not restricted to `np.float64` only

Please let me know if this looks right.

pykokkos/lib/ufuncs.py
ShahzadUmair: Oh yes, you are right. I will update the interface

pykokkos/interface/views.py
ShahzadUmair: I assumed that an `asarray` function can get a `list` as input as well (we are sending a list to asarray fn at 1 place in naive bayes example). Since lists are not hashable I believe, this code breaks with the following error

``` shell
File "/pykokkos1/pykokkos/interface/views.py", line 570, in asarray
    if obj in {pk.e, pk.pi, pk.inf, pk.nan}:
TypeError: unhashable type: 'list'
```

pykokkos/lib/ufuncs.py
ShahzadUmair: This is tied to fixing the `asarray` function. I am looking at fixing. Once we have a working `asarray` function, we should be able to remove this.

pykokkos/lib/ufuncs.py
ShahzadUmair: I think numpy does not do this. Since we do not have a mechanism in place to verify broadcasting, this will work for now assuming the inputs can be broadcasted.
Another option could be to make separate workunits for all the different combinations of inputs. Please let me know if you think we should go down that route instead.

pykokkos/lib/ufuncs.py
ShahzadUmair: Renamed it to `add_2d...`. Used modulo for the same reason here to prevent duplication of workunits for different combinations of inputs.

pykokkos/interface/views.py
NaderAlAwar: As discussed, `not hasattr(N, "__len__")` might make more sense here for now

pykokkos/interface/views.py
```diff
@@ -503,8 +516,10 @@ def from_numpy(array: np.ndarray, space: Optional[MemorySpace] = None, layout: O
     else:
         ret_list = list((array.shape))
 
-
-    return View(ret_list, dtype, space=space, trait=Trait.Unmanaged, array=array, layout=layout)
+    # Temporary Fix. Will revert when fixed
+    view = pk.View(ret_list, dtype)
+    view[:] = array
+    return view
```
NaderAlAwar, line 522: As discussed, create a custom `asarray` in GuassianNB, and use it where necessary

pykokkos/lib/ufuncs.py
```diff
@@ -584,6 +632,11 @@ def power_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D
 def power_impl_1d_float(tid: int, viewA: pk.View1D[pk.float], viewB: pk.View1D[pk.float], out: pk.View1D[pk.float]):
     out[tid] = pow(viewA[tid], viewB[tid])
 
+@pk.workunit
+def power_num_impl_2d_double(tid: int, viewA: pk.View2D[pk.double], viewB: pk.View1D[pk.double], out: pk.View2D[pk.double]):
```
NaderAlAwar, line 636: Make naming consistent with other workunits (remove num)

pykokkos/lib/util.py
NaderAlAwar: this might not be needed anymore since we support comparing to scalars now

NaderAlAwar (approved):

PR #90, opened 2022-09-26
ENH: support result_type

tylerjereddy:
* add changes needed to support the array API data type function `result_type()`, which is tested by
`array_api_tests/test_data_type_functions.py::test_result_type` and described in the standard at:
  - https://data-apis.org/array-api/latest/API_specification/generated/signatures.data_type_functions.result_type.html?highlight=result_type
  - https://data-apis.org/array-api/latest/API_specification/type_promotion.html#type-promotion

* surprisingly, this relatively simple draft logic already passes the array API test, but I'm certain this is far too simple to handle all cases for us beyond the simplest ones; still, it is a start on our type promotion rules

* I don't think this will work with views yet (or what we'll eventually call arrays), but it seems to pass the dtype tests used by the conformance suite at least

### DIFF
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -45,4 +45,4 @@ jobs:
           pip install -r requirements.txt
           export ARRAY_API_TESTS_MODULE=pykokkos
           # only run a subset of the conformance tests to get started
-          pytest array_api_tests/meta/test_broadcasting.py array_api_tests/meta/test_equality_mapping.py array_api_tests/meta/test_signatures.py array_api_tests/meta/test_special_cases.py array_api_tests/test_constants.py array_api_tests/meta/test_utils.py array_api_tests/test_creation_functions.py::test_ones array_api_tests/test_creation_functions.py::test_ones_like
+          pytest array_api_tests/meta/test_broadcasting.py array_api_tests/meta/test_equality_mapping.py array_api_tests/meta/test_signatures.py array_api_tests/meta/test_special_cases.py array_api_tests/test_constants.py array_api_tests/meta/test_utils.py array_api_tests/test_creation_functions.py::test_ones array_api_tests/test_creation_functions.py::test_ones_like array_api_tests/test_data_type_functions.py::test_result_type
--- a/pykokkos/interface/views.py
+++ b/pykokkos/interface/views.py
@@ -577,6 +577,60 @@ def asarray(obj, /, *, dtype=None, device=None, copy=None):
     return ret
 
 
+def result_type(*arrays_and_dtypes):
+    """
+    Returns the dtype that results from applying the type promotion rules to the arguments.
+
+    Parameters
+        arrays_and_dtypes (Union[array, dtype]) – an arbitrary number of input arrays and/or dtypes.
+
+    Returns
+        out (dtype) – the dtype resulting from an operation involving the input arrays and dtypes.
+
+    """
+    # TODO: we'll probably want objects for "categories
+    # of types" to simplify the logic here eventually..
+    types_seen = []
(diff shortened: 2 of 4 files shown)


.github/workflows/array_api.yml
```diff
@@ -45,4 +45,4 @@ jobs:
           pip install -r requirements.txt
           export ARRAY_API_TESTS_MODULE=pykokkos
           # only run a subset of the conformance tests to get started
-          pytest array_api_tests/meta/test_broadcasting.py array_api_tests/meta/test_equality_mapping.py array_api_tests/meta/test_signatures.py array_api_tests/meta/test_special_cases.py array_api_tests/test_constants.py array_api_tests/meta/test_utils.py array_api_tests/test_creation_functions.py::test_ones array_api_tests/test_creation_functions.py::test_ones_like
+          pytest array_api_tests/meta/test_broadcasting.py array_api_tests/meta/test_equality_mapping.py array_api_tests/meta/test_signatures.py array_api_tests/meta/test_special_cases.py array_api_tests/test_constants.py array_api_tests/meta/test_utils.py array_api_tests/test_creation_functions.py::test_ones array_api_tests/test_creation_functions.py::test_ones_like array_api_tests/test_data_type_functions.py::test_result_type
```
tylerjereddy, line 48: We may want to try splitting this over several lines before it gets too silly, although GitHub still does "ok" to highlight the change.

pykokkos/interface/views.py
```diff
@@ -577,6 +577,60 @@ def asarray(obj, /, *, dtype=None, device=None, copy=None):
...
+    """
+    Returns the dtype that results from applying the type promotion rules to the arguments.
+
+    Parameters
+        arrays_and_dtypes (Union[array, dtype]) – an arbitrary number of input arrays and/or dtypes.
+
+    Returns
+        out (dtype) – the dtype resulting from an operation involving the input arrays and dtypes.
+
+    """
```
tylerjereddy, line 590: Not exactly `numpydoc` compliant docstring, could fix that.

We don't handle mixed types yet--the standard gives us some flexibility on those, which is why we can get away with just the simple same-category logic for now. For example, it is up to us to decide how to handle a mixture of float and int types.

NaderAlAwar: I think we use different formats in different parts of the project, but I think for now you should use the format we use in this file. Also I think you're missing the type annotation for the return type

pykokkos/interface/views.py
```diff
@@ -577,6 +577,60 @@ def asarray(obj, /, *, dtype=None, device=None, copy=None):
...
+            if pk.iinfo(dtype).max > pk.iinfo(largest_int).max:
+                largest_int = dtype
+        return largest_int
+    elif float_types_seen and (not uint_types_seen) and (not int_types_seen):
+        largest_float = float_types_seen[0]
+        for dtype in float_types_seen[1:]:
+            if pk.finfo(dtype).max > pk.finfo(largest_float).max:
+                largest_float = dtype
+        return largest_float
+
```
NaderAlAwar, line 633: A few comments here:
1. Since we're returning in each body of the `if` and `elif` blocks, you can replace the `elif`s with `if`s.
2. Since the logic is identical in each `elif` block, I think you should extract the logic in each to a separate function, something like `get_largest_type(types_seen)` or something.
3. You're probably still working on this, but it seems nothing is returned if none of the conditions are met?
4. Can you add a unit test for this?

tylerjereddy:
I think I addressed most of the comments in the latest commit--see the commit message for details. I'm not sure I agreed with all the suggestions, but pretty sure I made them all anyway.

NaderAlAwar (approved):

PR #93, opened 2022-09-28
Enable usage of multiple GPUs in PyKokkos

NaderAlAwar:
This PR enables using multiple CUDA GPUs in PyKokkos in the same process. It depends on another pykokkos-base https://github.com/kokkos/pykokkos-base/pull/48 which makes multiple copies of libpykokkos, libkokkoscore, and libkokkoscontainers. PyKokkos will then link kernels compiled for CUDA to a different set of libraries depending on the GPU specified by the user.

The concept of Multi-GPU Kokkos is based on William Ruys's work [here](https://github.com/ut-parla/Parla.py/tree/main/examples/kokkos)

### DIFF
--- a/examples/pykokkos/multi_gpu.py
+++ b/examples/pykokkos/multi_gpu.py
@@ -1,0 +1,60 @@
+import pykokkos as pk
+
+import numpy as np
+import cupy as cp
+
+pk.set_default_space(pk.Cuda)
+
+size = 10000
+
+pk.set_device_id(0)
+cp_arr_0 = cp.arange(size).astype(np.int32)
+
+pk.set_device_id(1)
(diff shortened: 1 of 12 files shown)


JBludau:
hmm, this might be a bit late, but libs downstream could run into problems with this, as the initialization state of kokkos is tied to the lib. I might be overthinking this, but we could get difficult errors if other libs link against the wrong kokkos lib ... basically they all would need to replicate what we are doing here in order to extend anything in pykokkos/pykokkos-base (e.g. define Views of their own type). Furthermore, I think we would run into problems when using it in combination with a C++ program using Kokkos as the C++ side would not have a proper lib so link against?

PR #95, opened 2022-09-28
ENH: log2 to API standard

tylerjereddy:
* adjustments needed to pass `test_log2()` in the array API conformance suite; turn that test on in CI now

* this is quite similar to gh-92


pykokkos/lib/ufuncs.py
```diff
@@ -157,11 +169,24 @@ def log2(view):
...
-    if str(view.dtype) == "DataType.double":
-        pk.parallel_for(view.shape[0], log2_impl_1d_double, view=view)
-    elif str(view.dtype) == "DataType.float":
-        pk.parallel_for(view.shape[0], log2_impl_1d_float, view=view)
-    return view
+    out = pk.View(view.shape, view.dtype)
+    if "double" in str(view.dtype) or "float64" in str(view.dtype):
+        if view.shape == ():
+            # NOTE: is this really worth sending to a kernel?
+            pk.parallel_for(1, log2_impl_1d_double, view=view, out=out)
```
NaderAlAwar, line 176: For another ufunc that Umair worked on (I forget which one exactly), I asked him to simply call the `math.func()` on the scalar. You could do that here too if you think it's better.

NaderAlAwar (approved):

pykokkos/lib/ufuncs.py
tylerjereddy: `math.log2(view)` doesn't work here because it raises `ValueError: math domain error` when the input is `0`. I think NumPy returns `-inf` and only raises a warning. Perhaps C++/kokkos are similar to NumPy in that regard.

One other thing to think about is 0-d vs. scalar input, and if we want to return a "Python scalar" vs. a "pykokkos scalar." And what would a "pykokkos scalar" even mean? NumPy scalars can be confusing, they often print like regular Python scalars, but they are not the same thing (for example, a NumPy scalar integer wouldn't have the arbitrary precision/size of a pure Python integer [bignums]).

tylerjereddy:
I rebased/updated--we'll see if the CI agrees with my local testing, which seemed "ok" at least.

tylerjereddy:
@NaderAlAwar are the failures here related to the changes in pykokkos-base?

NaderAlAwar:
Yeah they are. Now that #109 is merged it should work once you rebase.

pykokkos/lib/ufuncs.py
NaderAlAwar: Does the array API say anything about 0-d and scalar inputs?

PR #97, opened 2022-09-28
ENH: isfinite() to API std

tylerjereddy:
* changes needed to pass `test_isfinite()` in the array API conformance suite, and turn this test on in CI

* this required me to write substantially more workunits/kernels for that ufunc, and to fill
in some more gaps in the type info system, including the scalar `int` conversion for single-element views


tylerjereddy:
I think the array API test suite is starting to share my viewpoint on the compiler performance:

> hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: Only produced 7 valid examples in 32.11 seconds

NaderAlAwar (approved):

PR #98, opened 2022-09-29
WIP, MAINT: pre-compile ufuncs

tylerjereddy:
* as discussed briefly on Slack, this is a draft branch for precompiling a subset of the ufuncs
to avoid JIT slowdown errors like the one
described here for the `hypothesis`-driven array
API conformance test suite:
https://github.com/kokkos/pykokkos/pull/97#issuecomment-1261624809

* for now, this focuses on `pk.double` type in a single dimension, simply because all ufuncs should support that for pre-compilation at this time (we can broaden the dim/type pre-compile as ufuncs gain more flexibile type/dim support)

* one major issue I see here is that I don't think the pre-compilation will actually do anything useful yet because of how the `pk_cpp` directory structures work on a per-module non-shared hierarchy basis:

- pre-compilation module `pk_cpp` hierarchy:
image
- `pytest` `pk_cpp` hierarchy:
image

### DIFF
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -49,6 +49,6 @@ jobs:
           # to circumvent the currently slow performance of
           # JIT compile/link, which can otherwise cause issues
           # for hypothesis-driven test case generation
-          python $GITHUB_WORKSPACE/pre_compile_tools/pre_compile_ufuncs.py
+          pytest $GITHUB_WORKSPACE/pre_compile_tools/pre_compile_ufuncs.py -s
           # only run a subset of the conformance tests to get started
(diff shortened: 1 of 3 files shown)


tylerjereddy:
Ok, with the latest changes here you can now do:

- `pytest ~/github_projects/pykokkos/pre_compile_tools/pre_compile_ufuncs.py -s` to pre-compile most ufuncs in about 5 minutes (the new infra can use `pytest` directly like this so that `pk_cpp` directory structure remains the same when you use `pytest` again for array API suite)
- then running the array API suite in the same directory has its runtime cut from 59 seconds to 23.7 seconds, which is better than 2X speedup

Note that a bunch of ufunc error handling additions were needed so that we can iterate over various types and shapes to force broad ufunc compilation.

I suppose this is close to what we discussed on Slack, though should only be treated as a temporary workaround for smoother JIT compilation and/or smoother pre-compilation infrastucture that doesn't require explicit calls like this.

I also cherry-picked the small fix from gh-99.

Issue #65, opened 2022-08-26
ENH: array API "namespace"


tylerjereddy:
When running the array API testsuite with `pykokkos`, one can see this warning: `hypothesis/extra/array_api.py:789: HypothesisWarning: Could not determine whether module pykokkos is an Array API library`.

If you look at the related code block in `hypothesis` testing library proper:

```python
    try:
        array = xp.zeros(1)
        array.__array_namespace__()
    except Exception:
        warn(
            f"Could not determine whether module {xp.__name__} is an Array API library",
            HypothesisWarning,
        )   
```

For details on the requirements for this `__array_namespace__()` method see:
- https://data-apis.org/array-api/latest/purpose_and_scope.html?highlight=array_namespace#how-to-adopt-this-api
- https://data-apis.org/array-api/latest/API_specification/generated/signatures.array_object.array.__array_namespace__.html#signatures.array_object.array.__array_namespace__

PR #99, opened 2022-09-29
CI: pin hypothesis

tylerjereddy:
* pin `hypothesis` to temporarily avoid https://github.com/kokkos/pykokkos/issues/65#issuecomment-1262840467

until we have a proper implementation of that


ShahzadUmair:
Just wondering if there is any benefit of pinning hypothesis version vs adding the `__array_api_version__` to our init file?

tylerjereddy:
Benefit is the simplicity of the fix for code review until the implementation is done properly.

NaderAlAwar (approved):

PR #101, opened 2022-09-30
added compiler_path when using env variable to point to kokkos install

JBludau:
The tuple the function should return should contain 3 paths but never did when pykokkos is configured to use an external Kokkos lib via `PK_KOKKOS_LIB_PATH`. Came up because ArborX uses Kokkos 3.6.01


NaderAlAwar (approved):
Good catch, thanks!

PR #102, opened 2022-09-30
Make pykokkos export bindings for the generated functor

JBludau:
From a discussion I had with @NaderAlAwar about extending the interoperability with C++. The main reasoning was the following:
If we let people write their functors in python and expose the generated C++ functors (incl. bindings) we enable people to use pybind for generating wrappers. Anyone who uses pykokkos for creating a binding to a C++ library that uses kokkos can use the functors as arguments to calls to his C++ library and then bind them. Basically we allow people to export functors (and thus algorithms/logic) into bindings of their own C++ libs.

Some considerations:

- Our functor requires every argument to be default constructible ... we should document that. We would not necessarily have to allow this but that is how it is at the moment.
- Thus it is no problem to also allow the generation of a default constructor which enables us to generate python bindings for the class. 
- Next step would be generate a cast operator from the python into the c++ instance so we do not copy data every time we switch to C++

### DIFF
--- a/pykokkos/core/translators/bindings.py
+++ b/pykokkos/core/translators/bindings.py
@@ -380,9 +380,35 @@ def generate_kernel(
 
     return kernel
 
-def bind_wrappers(module: str, wrappers: List[str]) -> str:
+def bind_wrapper(variable: str, wrapper: str) -> str:
     """
     Generate the binding code for all wrappers
+
+    :returns: the binding code
+    """
+
+    binding = f"{variable}.def(\"{wrapper}\", &{wrapper});"
+
+    return binding
+
+def bind_functor(
+    variable: str,
+    functor: str) -> List[str]:
+    """
+    Generates the bindings for a functor.
+
+    :param variable: the name of the module variable
+    :param functor: the functor class name
+    :returns: the binding code
+    """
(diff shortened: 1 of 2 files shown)


pykokkos/core/translators/bindings.py
```diff
@@ -380,17 +380,46 @@ def generate_kernel(
...
+    :returns: the binding code
+    """
+
+    binding = f"{variable}.def(\"{wrapper}\", &{wrapper});"
+
+    return binding
+
+def bind_functor(
+    variable: str,
+    functor: str) -> List[str]:
```
NaderAlAwar, line 396: I'm nitpicking here but can you put all the arguments on the same line?

NaderAlAwar (changes requested):
Looks good, thanks Jakob! Just the formatting change and I'll merge this

pykokkos/core/translators/bindings.py
JBludau: done in [4fba93](https://github.com/kokkos/pykokkos/pull/102/commits/4fba9361ee783695c9dd9bda0cc503729f21c72d)

NaderAlAwar (approved):

PR #107, opened 2022-10-10
Functor cast operators for interoperability with c++

JBludau:
This pr proposes custom "cast" functions to convert from a python object of the functor into the corresponding c++ functor. This allows other bindings that bind c++ libs that use kokkos to use pykokkos-generated functors in their bindings.

Why is it not a conventional "cast" operator but a free function?
A pybind11 cast operator would need to know how to construct the python object the functor wants to cast into. This is a problem as we do not know where the user defines the functor object and I did not want to import the main script into the c++ binding code just to know how to construct the functor object in c++.
This is solved by passing in a python object when casting from the c++ object to the python object. This leaves it to the user that calls the cast that the python object actually has all the members the cast assigns to ... the function has no guarantees or checks, as the interpreter should throw a runtime error if any of the assignments fails, which is as good as it gets. Both "casts" take care of the deep_copy and resize operations. This said, they are more like glorified copy operations

For this the following things were implemented:

- `functor_cast.py` script that generates the "cast" functions
- binding for the "cast" functions
- `compiler.py` and `cpp_setup.py` deal with the new file
- `translator.py` and `bindings.py` were refactored where necessary to avoid code repetition (this is just a proposal)
- Include guards for the functor.hpp and functor_cast.hpp file
- constructor for the c++ functor that does not get the rand stuff ... as we basically do an assignment which should not need it

### DIFF
--- a/pykokkos/core/translators/bindings.py
+++ b/pykokkos/core/translators/bindings.py
@@ -124,14 +124,14 @@ def get_device_views(members: PyKokkosMembers) -> Dict[str, str]:
     return {v.declname: f"pk_d_{v.declname}" for v in members.views \
             if members.views[v] is not None}
 
-def generate_functor_instance(functor: str, members: PyKokkosMembers, with_random_args: bool=True, functorExecSpace: Optional[str] = None) -> str:
+def generate_functor_instance(functor: str, members: PyKokkosMembers, with_random_args: bool=True, functor_exec_space: Optional[str] = None) -> str:
     """
     Generate the functor instance
 
     :param functor: the name of the functor
     :param members: an object containing the fields and views
     :param with_random_args: bool indicating if the constructor call should have the random args. default = True
-    :param functorExecSpace: optional parameter of type str that contains the ExecSpace template argument to the functor
+    :param functor_exec_space: optional parameter of type str that contains the ExecSpace template argument to the functor
     :returns: the source code for instantiating the functor
     """
 
@@ -140,11 +140,11 @@ def generate_functor_instance(functor: str, members: PyKokkosMembers, with_rando
     for f in members.fields:
         args.append(f.declname)
 
-    if functorExecSpace is None:
+    if functor_exec_space is None:
         exec_space: str = Keywords.DefaultExecSpace.value
         exec_space_instance: str = Keywords.DefaultExecSpaceInstance.value
     else:
-        exec_space: str = functorExecSpace
+        exec_space: str = functor_exec_space
         exec_space_instance: str = f"{exec_space}_instance"
 
     mirror_views: str = f"{exec_space} {exec_space_instance};"
--- a/pykokkos/core/translators/bindings.py
+++ b/pykokkos/core/translators/bindings.py
@@ -48,7 +48,7 @@ def get_view_memory_space(view_type: cppast.ClassType, location: str) -> str:
                 return f"Kokkos::{name}"
 
     if location == "functor":
-        return f"ExecutionSpace::memory_space"
+        return f"typename ExecSpace::memory_space"
     if location == "bindings":
         return Keywords.ArgMemSpace.value
 
@@ -383,7 +383,9 @@ def generate_kernel(
         acc = f"{return_type} {Keywords.Accumulator.value} = 0;"
 
     if members.has_real:
-        functor += f"<{real}>"
+        functor += f"<{Keywords.DefaultExecSpace.value},{real}>"
+    else:
+        functor += f"<{Keywords.DefaultExecSpace.value}>"
 
     instance: str = generate_functor_instance(functor, members)
     call: str = generate_call(operation, functor, members, tag, hierarchical)
--- a/pykokkos/core/translators/bindings.py
+++ b/pykokkos/core/translators/bindings.py
@@ -130,6 +130,8 @@ def generate_functor_instance(functor: str, members: PyKokkosMembers, with_rando
 
     :param functor: the name of the functor
     :param members: an object containing the fields and views
+    :param with_random_args: bool indicating if the constructor call should have the random args. default = True
+    :param functorExecSpace: optional parameter of type str that contains the ExecSpace template argument to the functor
     :returns: the source code for instantiating the functor
     """
 
--- a/pykokkos/core/translators/bindings.py
+++ b/pykokkos/core/translators/bindings.py
@@ -546,7 +546,9 @@ def bind_main_single(
         real = visitors_util.view_dtypes[precision.name].value
         wrapper_name += f"_{real}"
         kernel_name += f"_{real}"
-        functor += f"<{real}>"
+        functor += f"<{Keywords.DefaultExecSpace.value},{real}>"
+    else:
+        functor += f"<{Keywords.DefaultExecSpace.value}>"
 
     main: List[str] = translate_mains(source, functor, members, pk_import)
     params: Dict[str, str] = get_kernel_params(members, False, True, real)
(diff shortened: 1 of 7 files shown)


pykokkos/core/translators/bindings.py
```diff
@@ -156,26 +156,27 @@ def generate_functor_instance(functor: str, members: PyKokkosMembers) -> str:
     if len(args) == 0:
         args.append("0")
 
-    args.append(Keywords.RandPoolSeed.value)
-    args.append(Keywords.RandPoolNumStates.value)
+    if(with_random_args):
```
NaderAlAwar, line 159: You don't need the parentheses here, just `if with_random_args:`

pykokkos/core/translators/functor.py
```diff
@@ -147,6 +147,50 @@ def generate_constructor(
...
+
+    assignments.extend(generate_assignments(fields))
+    # skip subviews
+    assignments.extend(generate_assignments({v: views[v] for v in views if views[v]}))
+
+    body = cppast.CompoundStmt(assignments)
+
+    return cppast.ConstructorDecl("", name, params, body)
+
+
```
NaderAlAwar, line 193: I wonder if it's possible to refactor this so that we reuse `generate_constructor()`? Kind of like what you did in bindings.py for `generate_copy_back()`. Just add an extra boolean (e.g. `add_rand_params`) that guards the addition of the random pool parameters to the constructor.

pykokkos/core/translators/functor_cast.py
```diff
@@ -0,0 +1,121 @@
...
+            if d is DataType.real:
+                continue
+            source.append(generate_cast_from_object(functor,members,d))
+            source.append(generate_cast_to_object(functor,members,d))
+
+    else:
+        source.append(generate_cast_from_object(functor,members,None))
+        source.append(generate_cast_to_object(functor,members,None))
+
+    return source
```
NaderAlAwar, line 121: Maybe add short doc strings to briefly describe what each function does? It might seem redundant given the function names but in my experience it's helpful. You could copy some of the stuff you wrote in the PR description

NaderAlAwar (changes requested):
Looks good, some minor changes and I think this is ready to be merged

pykokkos/core/translators/functor_cast.py
JBludau: makes perfect sense. done in [8ee726](https://github.com/kokkos/pykokkos/pull/107/commits/8ee726edd636a9c5ee09c80baa2d7b73537ee79f)

pykokkos/core/translators/functor.py
JBludau: I was trying ... but the rationale for not doing it was that it would be a major change to the functors interface. In the state before this pr, the constructor of the functor was getting the rand arguments no matter what ... if they were not used in the constructor they would go out of scope. But everything in pykokkos now expects the functor to be constructible this way.
In c++ this kind of is a problem, as the random arguments alone make the functor dependent on the types that are the random_args even if they are never used in the constructor implementation. This basically ties the type of the functor to other types (seen by having to include `<Kokkos_Random.hpp>`) ... and any unnecessary dependencies in code will make future maintenance harder. Just think about that a change in `<Kokkos_Random.hpp>` might change the compile flags necessary ... and thus we would need to adapt both compilation and linking even though we might never use anything from the header. 
Thus I think in the long run, the constructor should only get the random args if it really uses them.
Even though this would be something I could do in this pr, it would be a different topic and I would rather do it in another pr (at best the two constructors that now do almost the same thing would be changed so that one uses the other so we do not repeat the implementation)
Furthermore, we might be interested in thinking about the functor constructors in general, as they require every member to be default constructible without actually needing to require this.
Additionally, we do a bunch of copies in the wrappers and the constructor that could be avoided. 

I guess the tl:dr is:
I think we should think about refactoring the functor creation in a broader context which deserves it's own pr. We could discuss this in the pykokkos meeting or in an issue

pykokkos/core/translators/bindings.py
JBludau: done in [c3f6f4](https://github.com/kokkos/pykokkos/pull/107/commits/c3f6f46b5e7f4b4e177211c99c7d0329e1708567)

pykokkos/core/translators/bindings.py
```diff
@@ -124,12 +124,14 @@ def get_device_views(members: PyKokkosMembers) -> Dict[str, str]:
...
 
-def generate_functor_instance(functor: str, members: PyKokkosMembers) -> str:
+def generate_functor_instance(functor: str, members: PyKokkosMembers, with_random_args: bool=True, functorExecSpace: Optional[str] = None) -> str:
     """
     Generate the functor instance
 
     :param functor: the name of the functor
     :param members: an object containing the fields and views
+    :param with_random_args: bool indicating if the constructor call should have the random args. default = True
+    :param functorExecSpace: optional parameter of type str that contains the ExecSpace template argument to the functor
```
NaderAlAwar, line 134: should be `functor_exec_space` to be consistent with our style

pykokkos/core/translators/bindings.py
```diff
@@ -138,44 +140,45 @@ def generate_functor_instance(functor: str, members: PyKokkosMembers) -> str:
...
     device_views: Dict[str, str] = get_device_views(members)
     for v, d_v in device_views.items():
         args.append(d_v)
-
-        view_type: cppast.ClassType = members.views[cppast.DeclRefExpr(v)]
-        if get_view_memory_space(view_type, "bindings") == Keywords.ArgMemSpace.value:
-            mirror_views += f"auto {d_v} = Kokkos::create_mirror_view_and_copy({exec_space_instance}, {v});"
-        else:
-            mirror_views += f"auto {d_v} = {v};"
+        mirror_views += f"auto {d_v} = Kokkos::create_mirror_view_and_copy({exec_space_instance}, {v});"
```
NaderAlAwar, line 155: What is the reason for this change? This helps avoid calling `create_mirror_view_and_copy` when the data is already in the right place, I believe it's used when passing Unmanaged views and for UVM space. Is it a no-op if the data is already on the device?

JBludau: so `create_mirror_view_and_copy` should take care of this, the documentation says:
If `std::is_same<typename Space::memory_space, typename ViewType::memory_space>::value` is false, creates a new `Kokkos::View` with the same layout and padding as `src` but with a device type of `Space::device_type` and conducts a `deep_copy` from `src` to the new view if one was created. Otherwise returns `src`.

I would just leave this to the Kokkos Core lib as adding another logic outside could break if anything changes within Kokkos. This is basically the reason we have these functions in the first place.
But apart from this: The cast operator uses this function to generate the constructor call. If it is dependent of the memory space, we would need a different version of the cast function for the case of the same memory space in `src` and `dst`.  That is why it made sense to do it in this pr (given that the kokkos function does the same logic we were repeating).

pykokkos/core/translators/bindings.py
JBludau: done in [75997c](https://github.com/kokkos/pykokkos/pull/107/commits/75997c5fb6fb490f52b97c33f92592f0d758ca11)

pykokkos/core/translators/bindings.py
JBludau: I talked to @NaderAlAwar about this and we decided to introduce a selector parameter which switches the behavior from the old one to a new version that always uses `create_mirror_view_and_copy`. [70f167](https://github.com/kokkos/pykokkos/pull/107/commits/70f167dd3f1747e61561150d94370bd5da041b9a)

NaderAlAwar (approved):

PR #108, opened 2022-10-10
PyKokkos: add support for HIPManagedSpace memory space

NaderAlAwar:
This adds HIPHostPinnedSpace support to pykokkos

### DIFF
--- a/pykokkos/core/cpp_setup.py
+++ b/pykokkos/core/cpp_setup.py
@@ -209,6 +209,9 @@ class CppSetup:
         if space is ExecutionSpace.Cuda:
             if enable_uvm:
                 view_space = "Kokkos::CudaUVMSpace"
+        if space is ExecutionSpace.HIP:
+            if enable_uvm:
+                view_space = "Kokkos::Experimental::HIPHostPinnedSpace"
 
         space_value: str
         if space.value == "HIP":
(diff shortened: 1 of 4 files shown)


JBludau:
There will be an alias `SharedHostPinnedSpace` and `SharedSpace` (for page migrating memory) in kokkos `4.0`. This together with the default spaces could become a backend portable solution to this. If we bump to Kokkos `3.7` there will be `HIPManagedSpace` which would be the equivalent to `CudaUVMSpace`. The HIP managed will have some quirks though, see https://github.com/kokkos/kokkos-tutorials/blob/main/Other/ReleaseBriefings/release-37.pdf

pykokkos/core/cpp_setup.py
```diff
@@ -209,6 +209,9 @@ def invoke_script(self, output_dir: Path, space: ExecutionSpace, enable_uvm: boo
         if space is ExecutionSpace.Cuda:
             if enable_uvm:
                 view_space = "Kokkos::CudaUVMSpace"
+        if space is ExecutionSpace.HIP:
+            if enable_uvm:
+                view_space = "Kokkos::Experimental::HIPHostPinnedSpace"
```
JBludau, line 214: hmm ... this will mean, that it will not be comparable to cuda ... should we hold it until we have `3.7`?

NaderAlAwar:
I wasn't aware of `HIPManagedSpace`. I'll work on resolving the 3.7 issues and use that instead

PR #109, opened 2022-10-11
Translator: add view hook template argument to view type for Kokkos 3.7

NaderAlAwar:
The `PK_NO_VIEW_HOOKS` environment variable can be set if an older Kokkos version is used

### DIFF
--- a/pykokkos/core/visitors/visitors_util.py
+++ b/pykokkos/core/visitors/visitors_util.py
@@ -1,4 +1,5 @@
 import ast
+import os
 import re
 import sys
 from typing import Dict, List, Optional, Set, Union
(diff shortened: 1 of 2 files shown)


pykokkos/core/visitors/visitors_util.py
```diff
@@ -336,6 +337,8 @@ def cpp_view_type(
         params_ordered.append(params["layout"])
     if "space" in params:
         params_ordered.append(params["space"])
+    if "PK_NO_VIEW_HOOKS" not in os.environ:
+        params_ordered.append("Kokkos::Experimental::DefaultViewHooks")
```
JBludau, line 341: Is this correct? I am somewhat confused by the combination of two negatives

NaderAlAwar: Yeah sorry the two negatives are confusing. Basically I wanted the default behavior to be the one that works with Kokkos 3.7, so the user needs to explicitly set the environment variable disabling them. So if that environment variable is not set, we want to add a hook.

JBludau: ok, if it does not work with pre 3.7, then why not name it more transparent ... e.g. `PK_KOKKOS_INTERFACE=3.6`? this would allow to extend it if the same problem appears another time.
Or we check the Kokkos version in the kokkos lib path (it will be somewhere in the cmake config that is generated)

PR #110, opened 2022-10-12
Revert "Make pykokkos export bindings for the generated functor"

NaderAlAwar:
Reverts kokkos/pykokkos#102

### DIFF
--- a/pykokkos/core/translators/bindings.py
+++ b/pykokkos/core/translators/bindings.py
@@ -380,33 +380,9 @@ def generate_kernel(
 
     return kernel
 
-def bind_wrapper(variable: str, wrapper: str) -> str:
+def bind_wrappers(module: str, wrappers: List[str]) -> str:
     """
-    Generate the binding code for a wrapper
-
-    :returns: the binding code
-    """
-
-    binding = f"{variable}.def(\"{wrapper}\", &{wrapper});"
-
-    return binding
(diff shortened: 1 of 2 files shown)


JBludau (approved):
Reasoning is the following: The definitions of the functor in the bindings (due to its global visibility) clashes if it is compiled for execution spaces. This could be addressed by making the functor a template of the execution space and only bind it once per execution space. Nevertheless, if we compile for multiple gpus this won't be enough as the execution space is tied to the gpu it is supposed to run on. As the functor contains information about the execution space (which is the same c++ type no matter which of the gpus it uses in the end but does not work for all the gpus) it will create functor bindings that only work for one of the used gpus.
This could be addressed with `module_local()` and binding the functor in every module for every gpu but if the bound functor is used in python, it is not one class but as many as we have gpus...and the user would need to manage this. 
We decided to revert #102 and change the "cast" operations in #107 to be only visible on the c++ side. This forces a copy via the cast operator every time the functor is passed from python to c++ but is more in line with what pykokkos is doing at the moment (always creating a local c++ copy of the functor from the python functor). Furthermore, it completely avoids the issues with multi gpu usage, as nothing that actually contains an executionspace is exposed on the python side but all things that point to the underlying kokkos library (and thus the respective gpu) is implicitly module local

PR #115, opened 2022-10-20
forward cpp_type from pykokkos_base to pykokkos

JBludau:
Complement to https://github.com/kokkos/pykokkos-base/pull/51 in pykokkos-base which makes the cpp_type accessible in the `View` class


JBludau:
closing since there is no progress and nothing ended up using it

PR #120, opened 2022-10-28
ENH: sign() to API std

tylerjereddy:
* adjust the `sign()` ufunc to pass the matching array API standard test, which has now been turned on in CI

* I think the array stuff is higher priority, but waiting to see what we decide to do with `bool` first over in `pykokkos-base` for that


pykokkos/lib/ufuncs.py
```diff
@@ -367,37 +367,396 @@ def log1p(view):
...
-    if len(view.shape) > 1:
-        raise NotImplementedError("only 1D views currently supported for sign() ufunc.")
-    if str(view.dtype) == "DataType.double":
-        pk.parallel_for(view.shape[0], sign_impl_1d_double, view=view)
-    elif str(view.dtype) == "DataType.float":
-        pk.parallel_for(view.shape[0], sign_impl_1d_float, view=view)
-    return view
+    out = pk.View(view.shape, view.dtype)
+    if len(view.shape) > 2:
+        raise NotImplementedError("only 2D views currently supported for sign() ufunc.")
```
tylerjereddy, line 638: should maybe be "up to 2D.."

NaderAlAwar: Do you want to modify this? I can merge it before it finishes rerunning the CI. I'm fine with it either way

NaderAlAwar (approved):

pykokkos/lib/ufuncs.py
tylerjereddy: I pushed in the string fixup with a commit message that disables running the CI again.

JBludau:
have you looked how this performs on gpus? might be bad but also we might not care for this function

NaderAlAwar:
We haven't looked at the performance of the ufuncs yet, we're doing this for conformance with the array API.

PR #123, opened 2022-11-02
Logistic Regression - Pykokkos example

ShahzadUmair:
This PR contains the basic working scikit learn's version of logistic regression.

We havent worked on the nested functions in this PR e.g. `sklearn.utils.validation.check_is_fitted` and a few others. As of now we just convert the output of these functions from numpy to pykokkos using `asarray`.

I am working on figuring out if we can get any significant gains from copying scikit learns internal functions and making them compatible with pykokkos.

This PR also fixes the mypy version to `0.981` because in the latest version mypy is testing some files which are in the ignore list (also pointed out [here](https://github.com/kokkos/pykokkos/pull/128#issuecomment-1326718976)).

### DIFF
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade numpy mypy cmake pytest pybind11 scikit-build patchelf tqdm
+          python -m pip install --upgrade numpy mypy==0.981 cmake pytest pybind11 scikit-build patchelf tqdm
       - name: Install pykokkos-base
         run: |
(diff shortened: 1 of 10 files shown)


JBludau:
this looks a little large to be reviewed ... is the `LR.py` copied from somewhere? If so we should mark it as such to give credit where it is due

NaderAlAwar:
Yeah it's the scikit-learn implementation of logistic regression using the pykokkos ufuncs. @ShahzadUmair could you please add the license information like we did for naive bayes? I'll review this shortly

NaderAlAwar (changes requested):
Minor change needed

tylerjereddy:
I don't think the `mypy` pin should have been needed since gh-129 was merged first.

PR #124, opened 2022-11-03
ENH: more ufuncs to API std

tylerjereddy:
* this uses some templating code to write 4 new API conformant ufuncs: `round`, `trunc`, `ceil`, and `floor`. (see: https://github.com/tylerjereddy/pykokkos_ufunc_generator )

* note that because these are automatically generated they have some verbosity/lack of docstring quirks, bit of an experiment..

* switch to the latest tagged release of the array API test suite rather than some random hash

* strangely, the API tests added here don't actually seem to need to compile the workunits--they may focus on the special use case for intergers I suppose; we may want to supplement with a few floating point examples perhaps; see also related discussion in
https://github.com/data-apis/array-api-tests/issues/154


pykokkos/lib/ufuncs.py
```diff
@@ -3009,3 +3009,307 @@ def isfinite(view):
...
+        out[tid][i] = floor(view[tid][i])
+
+
+@pk.workunit
+def floor_impl_3d_float(tid: int, view: pk.View3D[pk.float], out: pk.View3D[pk.float]):
+    for i in range(view.extent(1)):
+        for j in range(view.extent(2)):
+            out[tid][i][j] = floor(view[tid][i][j])
+
+def floor(view):
```
tylerjereddy, line 3275: lack of docstring is just the result of me not worrying about the docstrings in the auto-templating code I linked... we could add some into the template

There's also the question of whether it makes sense to commit auto-generated code vs. the stuff that generates it; conversely, committing to a templating engine could be annoying as well. Certainly, NumPy used its own custom templating language for years for generating the ufunc inner loops. I think they're starting to transition to C++ these days, at least in some limited areas.

pykokkos/lib/ufuncs.py
```diff
@@ -3009,3 +3009,307 @@ def isfinite(view):
...
+def ceil_impl_2d_float(tid: int, view: pk.View2D[pk.float], out: pk.View2D[pk.float]):
+    for i in range(view.extent(1)):
+        out[tid][i] = ceil(view[tid][i])
+
+
+@pk.workunit
+def ceil_impl_3d_float(tid: int, view: pk.View3D[pk.float], out: pk.View3D[pk.float]):
+    for i in range(view.extent(1)):
+        for j in range(view.extent(2)):
+            out[tid][i][j] = ceil(view[tid][i][j])
```
tylerjereddy, line 3197: Is this right? I'm afraid we may have to add some custom tests to supplement b/c I'm not seeing `pk_cpp` generated by test suite for `ceil`, `floor`, `round`, `trunc`..

pykokkos/lib/ufuncs.py
```diff
@@ -3009,3 +3009,307 @@ def isfinite(view):
...
+def round_impl_2d_float(tid: int, view: pk.View2D[pk.float], out: pk.View2D[pk.float]):
+    for i in range(view.extent(1)):
+        out[tid][i] = round(view[tid][i])
+
+
+@pk.workunit
+def round_impl_3d_float(tid: int, view: pk.View3D[pk.float], out: pk.View3D[pk.float]):
+    for i in range(view.extent(1)):
+        for j in range(view.extent(2)):
+            out[tid][i][j] = round(view[tid][i][j])
```
NaderAlAwar, line 3045: I don't think this will work, see https://github.com/kokkos/pykokkos/blob/a39282171630803067c8c2f13858e0a9beae2eb9/pykokkos/core/visitors/visitors_util.py#L70 for a list of functions that get mapped to their `cmath` counterparts

pykokkos/lib/ufuncs.py
NaderAlAwar: I think we should add the docstring in this PR. Regarding committing the templating engine, I would say not to do that for now. Since we're planning on getting rid of type annotations in the future, writing these kernels should be less tedious in the future.

pykokkos/lib/ufuncs.py
NaderAlAwar: Actually there is a `round` in cmath, it's just not part of this set. Could you add it here as well?

pykokkos/lib/ufuncs.py
NaderAlAwar: `round` should also work once we add it to the set of allowed math functions

NaderAlAwar (changes requested):
I think we just need to add `round` to the set of math functions and some short docstrings to the ufuncs

pykokkos/lib/ufuncs.py
```diff
@@ -3077,3 +3077,390 @@ def isfinite(view):
...
+
+    """
+    if "int" in str(view.dtype):
+        # special case defined in API std
+        return view
+    out = pk.View(view.shape, dtype=view.dtype)
+    if len(view.shape) > 3:
+        raise NotImplementedError("only up to 3D views currently supported for round() ufunc.")
+        
+    if "double" in str(view.dtype):
```
gliga, line 3143: Rather than having three blocks for three types, can we have one if statement and then call a function to which we pass three functions, something along these lines
```
if "double" in str(view.dtype):
   _parallel_for(round_impl_1d_double, round_impl_2d_double, round_impl_3d_double...)
elif "float64" in str(view.dtype): 
   ...
```

pykokkos/lib/ufuncs.py
```diff
@@ -3077,3 +3077,390 @@ def isfinite(view):
...
+        if view.shape == ():
+            pk.parallel_for(1, round_impl_1d_double, view=view, out=out)
+        elif len(view.shape) == 1:
+            pk.parallel_for(view.shape[0], round_impl_1d_double, view=view, out=out)
+        elif len(view.shape) == 2:
+            pk.parallel_for(view.shape[0], round_impl_2d_double, view=view, out=out)
+        elif len(view.shape) == 3:
+            pk.parallel_for(view.shape[0], round_impl_3d_double, view=view, out=out)
+
+    elif "float64" in str(view.dtype):
```
gliga, line 3153: instead of having `str(view.dtype)` each time, can we save this into a variable and use it when needed

pykokkos/lib/ufuncs.py
```diff
@@ -3077,3 +3077,390 @@ def isfinite(view):
...
+
+    """
+    if "int" in str(view.dtype):
+        # special case defined in API std
+        return view
+    out = pk.View(view.shape, dtype=view.dtype)
+    if len(view.shape) > 3:
+        raise NotImplementedError("only up to 3D views currently supported for trunc() ufunc.")
+        
+    if "double" in str(view.dtype):
```
gliga, line 3240: same comment as earlier: if we have `_parallel_for` we can save a lot of lines as it will be used here as well

pykokkos/lib/ufuncs.py
```diff
@@ -3077,3 +3077,390 @@ def isfinite(view):
...
+
+    """
+    if "int" in str(view.dtype):
+        # special case defined in API std
+        return view
+    out = pk.View(view.shape, dtype=view.dtype)
+    if len(view.shape) > 3:
+        raise NotImplementedError("only up to 3D views currently supported for floor() ufunc.")
+        
+    if "double" in str(view.dtype):
```
gliga, line 3434: same

gliga (commented):
Looks good overall, but it would be good we reduce some of those if blocks as suggested.

tylerjereddy:
We could reduce some code duplication, but the code is all generated from a template engine anyway, as noted above.

I suppose we could adjust the template engine, though I don't know how much easier things will be to read with more abstraction.

pykokkos/lib/ufuncs.py
tylerjereddy: How does `_parallel_for` know how to dispatch to the individual (1D vs. 2D vs. 3D) functions? How do you write a single `_parallel_for` that is shared between different ufuncs? I don't think we can easily template to the workunits without basically writing a bunch of nested logic again. And we need a separate `_parallel_for` for each type? I dunno, it seems  like the abstraction is risky on making things simpler because the workunits don't really have metadata we can hook into for dispatching.

tylerjereddy: I suppose we could write a dictionary with the workunits as values, and dispatch through well-named keys (types, and dimensions) perhaps. We might want to agree on how that would be designed first if we adopt something like that.

tylerjereddy:
I tried to write up a short proposal for improving our concision for kernel dispatch at gh-135. @gliga does that sound reasonable? I was thinking we might delay that slightly as a separate effort, since it would affect the dispatching behavior of all ufuncs I suspect.

gliga:
@tylerjereddy this PR is OK to go then.  I will check that separately and that can be revised in a diff PR. Thanks

tylerjereddy:
I rebased this on latest `develop`, and adjusted to use the new ufunc dispatching/templating system, so it is a bit more concise now. We can probably abstract even more I suppose.

NaderAlAwar (approved):

PR #134, opened 2022-12-13
ENH: DGEMM non-optimized

tylerjereddy:
* get some initial regression tests in place by adding a non-optimized DGEMM algorithm--specifically `Algorithm 1` from Kurzak et al. (2012)

* my main objective here is just to have some tests in place before we try to move on to more advanced algorithms or perhaps even autotuning? Not sure how far down that rabbit hole we plan to go just yet, and which knobs we can reasonably expect to turn from PyKokkos alone

@gliga @NaderAlAwar this starts what was briefly discussed in our email chain I think..

### DIFF
--- a/pykokkos/linalg/l3_blas.py
+++ b/pykokkos/linalg/l3_blas.py
@@ -5,7 +5,8 @@ import pykokkos as pk
 def dgemm(alpha: float,
           view_a,
           view_b,
-          beta: float = 1.0):
+          beta: float = 0.0,
+          view_c = None):
     """
     Double precision floating point genernal matrix multiplication (GEMM).
 
@@ -17,7 +18,8 @@ def dgemm(alpha: float,
              Shape (m, k)
     view_b : pykokkos view of type double
              Shape (k, n)
-    beta: float
+    beta: float, optional
+    view_c: pykokkos view of type double, optional
 
     Returns
     -------
@@ -48,6 +50,7 @@ def dgemm(alpha: float,
             for k in range(k_a):
                 subresult = view_a[m, k] * view_b[k, n] * alpha
                 C[m, n] += float(subresult) # type: ignore
-            C[m, n] *= beta # type: ignore
+            if view_c is not None:
+                C[m, n] += (view_c[m, n] * beta) # type: ignore
 
     return C
--- a/pykokkos/linalg/l3_blas.py
+++ b/pykokkos/linalg/l3_blas.py
@@ -1,0 +1,53 @@
+import pykokkos as pk
+
+# Level 3 BLAS functions
+
+def dgemm(alpha: float,
+          view_a,
+          view_b,
+          beta: float = 1.0):
+    """
+    Double precision floating point genernal matrix multiplication (GEMM).
+
+
+    Parameters
+    ----------
+    alpha: float
+    view_a : pykokkos view of type double
+             Shape (m, k)
+    view_b : pykokkos view of type double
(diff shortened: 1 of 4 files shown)


pykokkos/linalg/l3_blas.py
```diff
@@ -0,0 +1,53 @@
+import pykokkos as pk
+
+# Level 3 BLAS functions
+
+def dgemm(alpha: float,
+          view_a,
+          view_b,
+          beta: float = 1.0):
```
tylerjereddy, line 8: This should probably default to `0`, and we should add the `view_c` (optional output) argument per my misunderstanding of the usual interface in https://github.com/scipy/scipy/issues/17594.

tests/test_linalg.py
```diff
@@ -0,0 +1,100 @@
...
+def test_dgemm_vs_scipy(alpha,
+                        a,
+                        b,
+                        beta,
+                        expected_c):
+    # test against expected results from
+    # scipy.linalg.blas.dgemm
+    # NOTE: it seems that SciPy may ignore beta
+    # in its wrapper perhaps?
+    # see: https://github.com/scipy/scipy/issues/17594
```
tylerjereddy, line 81: As clarified over there, I suppose we'll need to add a `view_c` argument and test permutations with `beta` that way. At least, that seems to be how NETLIB LAPACK implementation works.

tylerjereddy:
Ok, I addressed some of my own comments following the upstream clarification of the LAPACK interface, and added a few more tests that use `beta` and `C` in combination, and they seem to match the SciPy `dgemm` behavior now. There are still a few more parameters that are used by SciPy/LAPACK for `dgemm` that I haven't added yet, not sure how important those are at this stage.

tylerjereddy:
This still needs some benchmarking code, maybe against the SciPy version, committed in here--currently slotted to work on that ~Wednesday I think.

tylerjereddy:
I pushed in some basic benchmarking infra for DGEMM vs. SciPy. Sample output on the terminal and plot from the script below. Obviously, OpenBLAS-driven SciPy is order of magnitude faster than pure Python.

`python benchmarks/dgemm_compare.py`

```
--------------------
system size: small
PyKokkos DGEMM execution time (s) for 50 repeats: 0.02863530395552516
SciPy DGEMM execution time (s) for 50 repeats: 9.210500866174698e-05
PyKokkos DGEMM timing is slower than SciPy with ratio: 310.90 fold
--------------------
--------------------
system size: medium
PyKokkos DGEMM execution time (s) for 50 repeats: 0.327494750963524
SciPy DGEMM execution time (s) for 50 repeats: 8.158502168953419e-05
PyKokkos DGEMM timing is slower than SciPy with ratio: 4014.15 fold
--------------------
--------------------
system size: large
PyKokkos DGEMM execution time (s) for 50 repeats: 5.225961541058496
SciPy DGEMM execution time (s) for 50 repeats: 0.0010483029764145613
PyKokkos DGEMM timing is slower than SciPy with ratio: 4985.16 fold
--------------------
df:
         PyKokkos     SciPy
small   0.028635  0.000092
medium  0.327495  0.000082
large   5.225962  0.001048
```

image

tylerjereddy:
Once, we're happy with this, I'd propose we merge and then I'll move forward with switching form "naive pure Python" to "naive workunit with OpenMP/CUDA," and after that we'll have to see how much performance gap remains re: need for autotuning, etc.

PR #137, opened 2022-12-25
ENH: acos implementation

tylerjereddy:
* add `acos()` ufunc implementation such that it passes the Python array API standard test, and turn that test on in our CI


JBludau (approved):
looks good to me

PR #138, opened 2022-12-26
ENH: acosh implementation

tylerjereddy:
* add `acosh()` ufunc implementation such that it passes the Python array API standard test, and turn that test on in our CI

This is quite similar to the plain `acos()` addition in gh-137, with a few function name/docstring adjustments.


JBludau (approved):
also looking good

PR #150, opened 2023-01-08
ENH: add less ufunc

tylerjereddy:
* draft the `less()` ufunc and turn its array API standard test on in CI

* some of the changes here are modified/copied from other recent binary ufunc pull requests, like the
broadcasting and type matching machinery; type matching in particular was expanded to include floating types, at least crudely

* there are probably some formatting improvements that could be made for some of the ufunc workunits added here--feel free to push those in--some of that work is basically just find/replace in text editor...


pykokkos/interface/views.py
```diff
@@ -785,3 +801,9 @@ class ScratchView7D(ScratchView, Generic[T]):
 
 class ScratchView8D(ScratchView, Generic[T]):
     pass
+
+
+def astype(view, dtype):
+    new_view = pk.View([*view.shape], dtype=dtype)
+    new_view[:] = view
+    return new_view
```
tylerjereddy, line 809: was needed by the array API testing machinery for `less()`

pykokkos/lib/ufunc_workunits.py
```diff
@@ -1,6 +1,513 @@
 import pykokkos as pk
 
 
+@pk.workunit
+def less_impl_5d_int8(tid: int,
+                              view1: pk.View5D[pk.int8],
+                              view2: pk.View5D[pk.int8],
+                              out: pk.View5D[pk.uint8]):
```
tylerjereddy, line 8: column-alignment could be cleaned up--feel free to help me with that... wasn't a high priority of course

pykokkos/lib/ufuncs.py
```diff
@@ -42,6 +42,51 @@ def _ufunc_kernel_dispatcher(tid,
...
+                effective_dtype = dtype1
+                view2_new = pk.View([*view2.shape], dtype=effective_dtype)
+                view2_new[:] = view2
+                view2 = view2_new
+            else:
+                effective_dtype = dtype2
+                view1_new = pk.View([*view1.shape], dtype=effective_dtype)
+                view1_new[:] = view1
+                view1 = view1_new
+    return view1, view2, effective_dtype
```
tylerjereddy, line 87: above two functions show up in some of my other binary ufunc branches/PRs, they may be slightly different here to accommodate expanding needs

PR #151, opened 2023-01-09
ENH: greater() to API standard

tylerjereddy:
* adjust the `greater()` ufunc to pass its array API standard test, and turn this test on in the CI

* this largely mirrors the changes in gh-150, except that we already had a draft implementation of `greater()` in place here


pykokkos/lib/ufunc_workunits.py
```diff
@@ -1,6 +1,513 @@
 import pykokkos as pk
 
 
+@pk.workunit
+def greater_impl_5d_int8(tid: int,
+                              view1: pk.View5D[pk.int8],
+                              view2: pk.View5D[pk.int8],
+                              out: pk.View5D[pk.uint8]):
```
tylerjereddy, line 8: usual comments about helping out with formatting apply -- feel free to push in here for all the copy-pasting we need to do b/c we don't have a templating lang

PR #152, opened 2023-01-10
ENH: equal() to API standard

tylerjereddy:
* adjust the `equal()` ufunc to pass its array API standard test, and turn this test on in the CI

* the changes include improved broadcasting and (type) casting support

* because equality testing is so common/rampant, this required a few more shims than I'd like...

Some warnings may show up in array API suite, feel free to help me with those:

```
array_api_tests/meta/test_utils.py::test_int_to_dtype
  /home/tyler/github_projects/pykokkos/pykokkos/interface/views.py:687: DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays.  The conversion of 1491047602 to int16 will fail in the future.
  For the old behavior, usually:
      np.array(value).astype(dtype)`
  will give the desired result (the cast overflows).
    arr = np.asarray(obj, dtype=dtype.np_equiv)
```


pykokkos/lib/ufunc_workunits.py
```diff
@@ -1,6 +1,563 @@
 import pykokkos as pk
 
 
+@pk.workunit
+def equal_impl_5d_int8(tid: int,
+                              view1: pk.View5D[pk.int8],
+                              view2: pk.View5D[pk.int8],
+                              out: pk.View5D[pk.uint8]):
```
tylerjereddy, line 8: as with several recent PRs, feel free to help with formatting for the large blocks of largely copy-paste templates of workunits..

NaderAlAwar (approved):

PR #160, opened 2023-01-17
ENH: multiply() to API std

tylerjereddy:
* bring the `multiply()` ufunc up to the array API standard and turn the matching test on in the CI

* similar to other recent binary ufunc PRs, with the usual drawbacks like broadcasting copies


pykokkos/lib/ufunc_workunits.py
```diff
@@ -1,6 +1,355 @@
 import pykokkos as pk
 
 
+@pk.workunit
+def multiply_impl_1d_double(tid: int, viewA: pk.View1D[pk.double], viewB: pk.View1D[pk.double], out: pk.View1D[pk.double], ):
+    out[tid] = viewA[tid] * viewB[tid % viewB.extent(0)]
```
tylerjereddy, line 6: the modulo operator is leftover from before here--it seems harmless in local testing but should probably be removed (though it may have some merit depending on how broadcasting evolves I suppose..)

PR #162, opened 2023-01-20
ENH: remainder() to API standard

tylerjereddy:
* draft the `remainder()` ufunc and turn its matching array API standad test on in the CI; many of the other changes reflect those in recent binary ufunc PRs

* note that the implementation of the remainder is actually rather non-trivial here; for reference, the reviewer may want to check the following, and perhaps also the NumPy inner loop implementation:
  - https://stackoverflow.com/a/1907585/2942522
  - https://data-apis.org/array-api/latest/API_specification/generated/array_api.remainder.html


tylerjereddy:
This one isn't ready yet--skipping the CI while I iterate on it. In particular, I think that for floating point type workunits I need to more closely obey the special cases noted at: https://data-apis.org/array-api/latest/API_specification/generated/array_api.remainder.html

That is indeed quite messy so will likely take me a while. It is no wonder modulus is not recommended for floating-point types.

PR #174, opened 2023-02-16
ENH: add full_like

tylerjereddy:
* add `full_like` creation function and turn its array API standard test on in the CI
* this borrows heavily from gh-152 and gh-141, out of necessity, so those should probably be reviewed/merged first, then carefully check the diff here after rebasing
* didn't add the docstring for `full_like` yet--feel free to help me get that in by adapting from https://data-apis.org/array-api/latest/API_specification/generated/array_api.full_like.html?highlight=full_like


NaderAlAwar (approved):

tylerjereddy:
Ah, gh-152 and gh-141 should have been reviewed/merged first, as noted above, because now when I modify those ufuncs the changes aren't guarded from regression. I'll try to clean that up in my next PR, but this could start getting complicated with features merged without their tests.

PR #181, opened 2023-03-02
ENH: allow type double scratch

### LINKED ISSUE #180: ENH: type support for scratch memory
While working on tiled matmul I noticed the following apparent limitation, which we should probably deal with for the DGEMM optimization exploration.

When adding this to a workunit, the error below shows up:

```python
scratch_mem: pk.ScratchView1D[double] = pk.ScratchView1D(team_member.team_scratch(0))
```

If we use `float` we're apparently fine though:

```python
scratch_mem: pk.ScratchView1D[float] = pk.ScratchView1D(team_member.team_scratch(0))
```

```
--------------------------------------------------------------------------------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------------------------------------------------------------------------------

Error on line 79 : Type not supported
DEBUG AST:
AnnAssign(
    target=Name(id='scratch_mem', ctx=Store()),
    annotation=Subscript(
        value=Attribute(
            value=Name(id='pk', ctx=Load()),
            attr='ScratchView1D',
            ctx=Load()),
        slice=Name(id='double', ctx=Load()),
        ctx=Load()),
    value=Call(
        func=Attribute(
            value=Name(id='pk', ctx=Load()),
            attr='ScratchView1D',
            ctx=Load()),
        args=[
            Call(
                func=Attribute(
                    value=Name(id='team_member', ctx=Load()),
                    attr='team_scratch',
                    ctx=Load()),
                args=[
                    Constant(value=0)],
                keywords=[])],
        keywords=[]),
    simple=1)
    scratch_mem: pk.ScratchView1D[double] = pk.ScratchView1D(team_member.team_scratch(0))
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Translation of <ast.FunctionDef object at 0x7fcfd60892a0> failed
```

tylerjereddy:
Fixes #180

* allow for scratch memory to be viewed as type `double`, including a regression test

* this is needed for shared memory usage in the DGEMM tiling algorithm (D = `double`)

* it turns out that `float` was aliased to `double`, which is incredibly confusing--I've adjusted such that `float` means `float` and `double` means `double`, which seems like the only sane design

### DIFF
--- a/pykokkos/core/visitors/visitors_util.py
+++ b/pykokkos/core/visitors/visitors_util.py
@@ -195,6 +195,17 @@ def get_type(annotation: Union[ast.Attribute, ast.Name, ast.Subscript], pk_impor
 
             return cppast.ClassType(type_name)
 
+    if isinstance(annotation, ast.Index):
+        # ast.Index has been deprecated since Python 3.9;
+        # this module attempts to shim around it, but we're
+        # still getting issues in gh-181 with Python 3.8
+
+        # we should probably drop support for Python 3.8 soon anyway
+        # per NEP29, but for now we attempt to handle ast.Index
+
+        # should convert to ast.Name:
+        annotation = annotation.value
+
     if isinstance(annotation, ast.Name):
         type_name: str = annotation.id
 
--- a/pykokkos/core/visitors/visitors_util.py
+++ b/pykokkos/core/visitors/visitors_util.py
@@ -15,7 +15,8 @@ def pretty_print(node):
 
 allowed_types: Dict[str, str] = {
     "int": "int",
-    "float": "double",
+    "float": "float",
+    "double": "double",
     "bool": "bool",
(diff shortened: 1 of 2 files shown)


tylerjereddy:
I pushed in a commit to attempt to deal with a CI failures with Python 3.8 related to `ast.Index` support. This took a few hours to debug, and we should probably just drop Python 3.8 support per NEP 29 anyway--I'm going to open an issue about that b/c maintenance burden is becoming non-trivial there.

tylerjereddy:
Why isn't the CI re-running.. I wonder if the Kokkos project is over-using some allocation or something?

tylerjereddy:
Ok, this is blocking the DGEMM tiling work so I'm going to self-merge with CI passing.

PR #198, opened 2023-05-10
Merge Develop to Main

NaderAlAwar:
I think for now we will not be using the develop branch as that has caused some confusion among users.


tylerjereddy (approved):
> I think for now we will not be using the develop branch as that has caused some confusion among users.

Sounds good to me, I think we test fairly carefully now too.

tylerjereddy:
One thing worth changing in this case is the GitHub configs for CI stuff from `develop` to `main` I think:

```yaml
on:
  push:
    branches:
      - develop
  pull_request:
    branches:
      - develop

```

PR #205, opened 2023-09-15
Abstract the from_cupy and from_numpy functions with a wrapper

Zhvhe:
Automatically convert cupy and numpy arrays (and other array like or convertabe objects) to a View without user input

### DIFF
--- a/benchmarks/dgemm_compare.py
+++ b/benchmarks/dgemm_compare.py
@@ -48,9 +48,9 @@ if __name__ == "__main__":
             b_new = b
             c_new = c
 
-        view_a = pk.from_array(a_new)
-        view_b = pk.from_array(b_new)
-        view_c = pk.from_array(c_new)
+        view_a = pk.array(a_new)
+        view_b = pk.array(b_new)
+        view_c = pk.array(c_new)
 
         pk_dgemm_time_sec = timeit.timeit("pk_dgemm(alpha, view_a, view_b, beta, view_c)",
                                           globals=globals(),
--- a/benchmarks/dgemm_compare.py
+++ b/benchmarks/dgemm_compare.py
@@ -48,9 +48,10 @@ if __name__ == "__main__":
             b_new = b
             c_new = c
 
-        view_a = pk.from_numpy(a_new)
-        view_b = pk.from_numpy(b_new)
-        view_c = pk.from_numpy(c_new)
+        view_a = pk.from_array(a_new)
+        view_b = pk.from_array(b_new)
+        view_c = pk.from_array(c_new)
+
         pk_dgemm_time_sec = timeit.timeit("pk_dgemm(alpha, view_a, view_b, beta, view_c)",
                                           globals=globals(),
                                           number=num_repeats)
--- a/examples/pykokkos/from_array.py
+++ b/examples/pykokkos/from_array.py
@@ -16,17 +16,23 @@ size = 10
 
 np_arr = np.zeros((size, 2)).astype(np.int32)
 cp_arr = cp.zeros((size, 2)).astype(np.int32)
+list_arr = [np.array([0, 0], dtype=np.int32)] * size
 
 print(cp_arr.flags)
 
 print(f"before {np_arr=}")
 print(f"before {cp_arr=}")
+print(f"before {list_arr=}")
 
 np_view = pk.from_numpy(np_arr)
 cp_view = pk.from_cupy(cp_arr)
+list_view = pk.from_array(list_arr)
 
 pk.parallel_for(pk.RangePolicy(pk.OpenMP, 0, size), addition_np, np_arr=np_view)
 pk.parallel_for(pk.RangePolicy(pk.Cuda, 0, size), addition_cp, cp_arr=cp_view)
+pk.parallel_for(pk.RangePolicy(pk.OpenMP, 0, size), addition_np, np_arr=list_view)
 
 print(f"after {np_arr=}")
 print(f"after {cp_arr=}")
+list_arr = [[(list_view[i][j], type(list_view[i][j])) for j in range(2)] for i in range(size)]
+print(f"after {list_arr=}")
--- a/examples/pykokkos/from_array.py
+++ b/examples/pykokkos/from_array.py
@@ -4,13 +4,13 @@ import pykokkos as pk
 
 @pk.workunit(np_arr = pk.ViewTypeInfo(space=pk.HostSpace))
 def addition_np(i: int, np_arr: pk.View2D[int]):
-    np_arr[i][0] += 1
-    np_arr[i][1] += 2
+    np_arr[i][0] += 1*i
+    np_arr[i][1] += 2*i
 
 @pk.workunit(cp_arr = pk.ViewTypeInfo(space=pk.CudaSpace, layout=pk.LayoutRight))
 def addition_cp(i: int, cp_arr: pk.View2D[int]):
-    cp_arr[i][0] += 1
-    cp_arr[i][1] += 2
+    cp_arr[i][0] += 1*i
+    cp_arr[i][1] += 2*i
 
 size = 10
 
@@ -24,9 +24,9 @@ print(f"before {np_arr=}")
 print(f"before {cp_arr=}")
 print(f"before {list_arr=}")
 
-np_view = pk.from_numpy(np_arr)
-cp_view = pk.from_cupy(cp_arr)
-list_view = pk.from_array(list_arr)
+np_view = pk.array(np_arr)
+cp_view = pk.array(cp_arr)
+list_view = pk.array(list_arr)
 
 pk.parallel_for(pk.RangePolicy(pk.OpenMP, 0, size), addition_np, np_arr=np_view)
 pk.parallel_for(pk.RangePolicy(pk.Cuda, 0, size), addition_cp, cp_arr=cp_view)
--- a/pykokkos/interface/__init__.py
+++ b/pykokkos/interface/__init__.py
@@ -49,8 +49,7 @@ from .views import (
     ScratchView, ScratchView1D, ScratchView2D,
     ScratchView3D, ScratchView4D, ScratchView5D,
     ScratchView6D, ScratchView7D, ScratchView8D,
-    from_cupy, from_numpy, from_array,
-    asarray, result_type,
+    array, asarray, result_type,
 )
 
 from .ext_module import compile_into_module
--- a/pykokkos/interface/__init__.py
+++ b/pykokkos/interface/__init__.py
@@ -49,7 +49,7 @@ from .views import (
     ScratchView, ScratchView1D, ScratchView2D,
     ScratchView3D, ScratchView4D, ScratchView5D,
     ScratchView6D, ScratchView7D, ScratchView8D,
-    from_cupy, from_numpy,
+    from_cupy, from_numpy, from_array,
     asarray, result_type,
 )
 
--- a/pykokkos/interface/views.py
+++ b/pykokkos/interface/views.py
@@ -15,8 +15,6 @@ import pykokkos as pk
 from pykokkos.bindings import kokkos
 import pykokkos.kokkos_manager as km
 
-import logging
-
 from .data_types import (
     DataType, DataTypeClass,
     real,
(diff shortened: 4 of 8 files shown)


benchmarks/dgemm_compare.py
```diff
@@ -48,9 +48,10 @@
             b_new = b
             c_new = c
 
-        view_a = pk.from_numpy(a_new)
-        view_b = pk.from_numpy(b_new)
-        view_c = pk.from_numpy(c_new)
+        view_a = pk.from_array(a_new)
```
gliga, line 51: Do we need from_array or can we just say from?

examples/pykokkos/from_array.py
```diff
@@ -16,17 +16,23 @@ def addition_cp(i: int, cp_arr: pk.View2D[int]):
 
 np_arr = np.zeros((size, 2)).astype(np.int32)
 cp_arr = cp.zeros((size, 2)).astype(np.int32)
+list_arr = [np.array([0, 0], dtype=np.int32)] * size
```
gliga, line 19: this is an example that we want to keep to show that explicit functions can be used, right?

If yes, we should have a corresponding example that looks more or less the same, but uses only from

pykokkos/interface/__init__.py
```diff
@@ -49,7 +49,7 @@
     ScratchView, ScratchView1D, ScratchView2D,
     ScratchView3D, ScratchView4D, ScratchView5D,
     ScratchView6D, ScratchView7D, ScratchView8D,
-    from_cupy, from_numpy,
+    from_cupy, from_numpy, from_array,
```
gliga, line 52: Do we want to even maintain explicit functions for diff arrays? @NaderAlAwar

pykokkos/interface/views.py
```diff
@@ -15,6 +15,8 @@
 from pykokkos.bindings import kokkos
 import pykokkos.kokkos_manager as km
 
+import logging
```
gliga, line 18: @NaderAlAwar can you confirm that you are fine that logging package is used for reporting here

pykokkos/interface/views.py
```diff
@@ -704,6 +706,53 @@ def from_cupy(array) -> ViewType:
...
+    """
+    Check if an object conforms to enough numpy array standards to be treated as an array
+    Using the function from_cupy() as a standard for bare minimum needed to
+    to convert any array into a numpy array
+
+    :param array: the array of unknown type
+    :returns: a true/false if object is an array
+    """
+    
+    required_attr = ["dtype", "data", "shape", "flags"]
```
gliga, line 719: probably should be global on the module level

examples/pykokkos/from_array.py
NaderAlAwar, line 19: When you initialize a list of objects like this all the elements of the list will reference the same object. Is this what you wanted here?

pykokkos/interface/__init__.py
NaderAlAwar: It might make sense to keep them internally but they should not be exposed to the user.

pykokkos/interface/views.py
```diff
@@ -704,6 +706,53 @@ def from_cupy(array) -> ViewType:
...
+    test_attr = dir(array)
+
+    if(not set(required_attr).issubset(set(test_attr))):
+      return False
+    
+    for d in required_attr:
```
gliga, line 725: this loop is confusing: if we enter we either return True or False, why do we have a loop? `return True` outside `for`

pykokkos/interface/views.py
```diff
@@ -704,6 +706,53 @@ def from_cupy(array) -> ViewType:
...
+    Create a PyKokkos View from a generic array
+
+    :param array: the array of unknown type
+    :param space: an optional argument for memory space (used by from_cupy)
+    :param layout: an optional argument for layout (used by from_cupy)
+    :returns: a PyKokkos View wrapping the array
+    """
+
+    if isinstance(array, np.ndarray):
+      return from_numpy(array, space, layout)
```
gliga, line 742: there is some problem with formatting in new code: old code appears to be using 4 spaces but new code is using 2; we should be consistent and use 4. I assume that the linter will catch this, but wanted to mention just in case

pykokkos/interface/views.py
```diff
@@ -704,6 +706,53 @@ def from_cupy(array) -> ViewType:
...
+      return from_cupy(array)
+
+    try:
+      # try converting the input array to numpy and using that route to convert
+      _array = np.asarray(array)
+    except Exception as e: 
+        logging.error("%s: Could not convert array to View", e)
+        raise ValueError
+
+    return from_numpy(_array, space, layout)
```
gliga, line 755: do we use `_` in variables at other places in a similar style?

tests/test_linalg.py
```diff
@@ -133,8 +133,8 @@ def test_dgemm_vs_scipy(alpha,
 
 def test_dgemm_input_handling():
     alpha = 1.0
-    view_a = pk.from_numpy(np.zeros((4, 3)))
-    view_b = pk.from_numpy(np.zeros((4, 3)))
+    view_a = pk.from_array(np.zeros((4, 3)))
+    view_b = pk.from_array(np.zeros((4, 3)))
```
gliga, line 137: these changes are good, but we should test with something else that implements the protocol

gliga (changes requested):
Overall these are good changes.
One question is if we want to keep those explicit conversions still available in the "public" API or we want to offer only one entry method `from`.

See also my individual comments about some formatting changes.

pykokkos/interface/views.py
```diff
@@ -704,6 +706,53 @@ def from_cupy(array) -> ViewType:
...
+    # numpy array sufficient to our purpose
+    if is_array(array):
+      return from_cupy(array)
+
+    try:
+      # try converting the input array to numpy and using that route to convert
+      _array = np.asarray(array)
+    except Exception as e: 
+        logging.error("%s: Could not convert array to View", e)
+        raise ValueError
```
NaderAlAwar, line 753: This isn't how I would do this. This doesn't really tell the user what went wrong. Maybe in this case just let numpy throw the exception directly.

pykokkos/interface/views.py
```diff
@@ -704,6 +706,53 @@ def from_cupy(array) -> ViewType:
...
+    :param layout: an optional argument for layout (used by from_cupy)
+    :returns: a PyKokkos View wrapping the array
+    """
+
+    if isinstance(array, np.ndarray):
+      return from_numpy(array, space, layout)
+    # test if the input array can duck-type to a
+    # numpy array sufficient to our purpose
+    if is_array(array):
+      return from_cupy(array)
```
NaderAlAwar, line 746: This logic feels a little weird to me. We're saying if it's a numpy array, call from numpy. But if it's an array, we call from_cupy? Why?

pykokkos/interface/views.py
NaderAlAwar: We wouldn't need this after removing the call to the logger later.

NaderAlAwar (changes requested):
Few things to look at:
1. The initialization of list_arr in from_array.py
2. The try except block in views.py
3. rename from_array to just array
I also don't fully understand the logic of calling from_numpy vs from_cupy in from_array

benchmarks/dgemm_compare.py
gliga: `array` is fine, but having one is way better than having three

examples/pykokkos/from_array.py
gliga: but we do not want explicit functions

pykokkos/interface/__init__.py
gliga: I like that way more than exposing each of them

pykokkos/interface/views.py
gliga: Ok, but that was not my questions: do we want to keep any logging in these files using the logging package.

benchmarks/dgemm_compare.py
Zhvhe: I'm fine with hiding all the rest so we only have one interface... just wasn't sure if we wanted to keep the backwards compatibility or not. I like array(), if we are hiding the others, I think it makes sense to be as numpy-like as possible. I only used from_array() to match the existing functions.

pykokkos/interface/__init__.py
Zhvhe: I can hide them! from_array() (or whatever we are calling it) is just a wrapper to both of them anyways...

pykokkos/interface/views.py
Zhvhe: Do we use it elsewhere? I basically stole this from the from_cupy() function and just rephrased a bit.

pykokkos/interface/views.py
Zhvhe: an attribute can be callable (a class function) or not (class variable). If the attribute happens to have the same name but is callable (as aoopsed to plain data)... I think the conversion would crash and burn.

pykokkos/interface/views.py
Zhvhe: Can fix!

pykokkos/interface/views.py
Zhvhe: the from_cupy function is actually a general function, not actually cupy specific. If the array is sufficiently numpy-like, without being numpy, the from_cupy function in theory can handle it.

Zhvhe: Yeah... the naming initial confused me too, had to go through the function comments to actually see what was going on. I think its ever actually used for cupy so in practice its fine but theoretically is can be used for other array types.

pykokkos/interface/views.py
Zhvhe: can do!

tests/test_linalg.py
Zhvhe: I can add a 1D array test? the 2D arrays convert badly with lists because the types get weird when you have lists of lists.

benchmarks/dgemm_compare.py
gliga: let's do `array` and hide others

pykokkos/interface/views.py
gliga: If a constant, no need to keep it a local var. I prefer we extract.

pykokkos/interface/views.py
gliga: My question is about the return statement inside the loop body (not outside)

tests/test_linalg.py
gliga: Let's start with 1D

pykokkos/interface/views.py
gliga: Please include necessary comments.

Zhvhe: I can rename from_cupy to from_array as that s more accurate?

tests/test_linalg.py
Zhvhe: got around the list of lists problems so did not write a 1D example here

examples/pykokkos/from_array.py
Zhvhe: Made the workload slightly more interesting to clarify what each thread was doing. Looks like it figured it out!

Zhvhe: Because of the numpy translation in the backend, you have to pull the data out of the view manually to return to a list format but it does seem to just make N copies when doing this: [ <init_object>] * N

pykokkos/interface/views.py
NaderAlAwar: Not sure, I would have to go over how they are currently used to see what changes are necessary

NaderAlAwar (approved):
Looks good to me

tests/test_views.py
```diff
@@ -298,6 +305,12 @@ def test_arrays(self):
             self.assertEqual(1, cp_arr[i][0])
             self.assertEqual(2, cp_arr[i][1])
 
+            self.assertEqual(1, list_view[i][0])
+            self.assertEqual(2, list_view[i][1])
+            self.assertEqual(1, list_arr[i][0])
+            self.assertEqual(2, list_arr[i][1])
+
```
gliga, line 312: why do we have these extra empty lines?

gliga (approved):
LG.

PR #206, opened 2023-09-20
Automatic Type Inferrence support

HannanNaeem:
Working on removing explicit type definition to make pykokkos more usable and "pythonic". This is the first step; where the main contribution is getting rid of the need to provide type annotations in the workunit definition

- Only supports standalone workunit format.
- Only supports parameters in workunit signatures NOT the body.
- Supports dynamically changing types for workunit.
- Adds dynamic inference for layout specifiers in workunit decorator.
- Expands on module/.so caching against data types of arguments the workunit was compiled with.
- Is backwards compatible; allows the user to override inference with explicitly providing type annotations.
- values passed to 'parallel_dispatch.py' functions are intercepted and passed for type inference.
- `interface/args_type_inference.py` contains all logic for type inference in `get_annotations` and helper functions.
- Moved Argument handling to 'args_type_inference.py` (Everything argument related is here now).
- Updated `Parser` (`parser.py`): added `fix_types`, `fix_view_layout`, and helper functions called there-in to modify the workunit AST and add the missing type annotations and layout decorators respectively.
- Replaced  examples in for standalone format with new syntax.
- Added tests for type inference logic in the test suite `test/test_typeinference`.


pykokkos/core/compiler.py
```diff
@@ -64,20 +64,30 @@ def compile_object(
         """
 
         metadata = module_setup.metadata
-
         hash: str = self.members_hash(metadata.path, metadata.name)
+        parser = self.get_parser(metadata.path)
+        entity: PyKokkosEntity = parser.get_entity(metadata.name)
+
+
```
NaderAlAwar, line 71: remove the extra new line

pykokkos/core/compiler.py
```diff
@@ -64,20 +64,30 @@ def compile_object(
...
-
         hash: str = self.members_hash(metadata.path, metadata.name)
+        parser = self.get_parser(metadata.path)
+        entity: PyKokkosEntity = parser.get_entity(metadata.name)
+
+        
         if self.is_compiled(module_setup.output_dir):
             if hash not in self.members: # True if pre-compiled
+                if updated_types is not None:
+                #* Fixing types
```
NaderAlAwar, line 75: This comment is not needed

pykokkos/core/compiler.py
```diff
@@ -64,20 +64,30 @@ def compile_object(
...
+                    parser.fix_types(entity, updated_types)
+                # before we hit extract members
```
NaderAlAwar, line 77: Also not needed

pykokkos/core/cpp_setup.py
```diff
@@ -371,5 +371,4 @@ def is_compiled(output_dir: Path) -> bool:
         :param output_dir: the directory containing the compiled entity
         :returns: true if compiled
         """
-
```
NaderAlAwar, line 374: Keep the newline here

pykokkos/core/compiler.py
```diff
@@ -64,20 +64,30 @@ def compile_object(
...
         metadata = module_setup.metadata
-
         hash: str = self.members_hash(metadata.path, metadata.name)
+        parser = self.get_parser(metadata.path)
+        entity: PyKokkosEntity = parser.get_entity(metadata.name)
+
+        
         if self.is_compiled(module_setup.output_dir):
             if hash not in self.members: # True if pre-compiled
+                if updated_types is not None:
```
NaderAlAwar, line 74: In what cases is `updated_types` expected to be `None`? Is it when the types are static?

pykokkos/core/module_setup.py
```diff
@@ -181,5 +181,4 @@ def is_compiled(self) -> bool:
         """
         Check if this module is compiled for its execution space
         """
-
```
NaderAlAwar, line 184: also keep the newline

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +124,165 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
 
+
+    def fix_types(self, entity: PyKokkosEntity, updated_types: UpdatedTypes) -> ast.AST:
+
+        '''
+        updatedTypes: object that contains info about inferred types
+
+        This method will walk the ast for a workunit and add any mission annotations
+        This method will also invoke fix_viewlayouts that will add missing decorators for user specified layouts
+        '''
```
NaderAlAwar, line 149: Look at the docstrings for the other functions and use the same format

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +124,165 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+        
+        check_entity: Callable[[ast.stmt], bool]
+        style: PyKokkosStyles = entity.style
+
+        # only supports standalone workunits, return the entity AST as it is
+        if style is not PyKokkosStyles.workunit or updated_types == None:
+            return entity.AST
```
NaderAlAwar, line 156: In what cases is this condition expected to be met? Are we calling this function from somewhere where the style might not be a workunit or `updated_types` might be None? I'd rather we have an assert statement here

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +124,165 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+        
+        check_entity = self.is_workunit
+
+        # For now, just so we can raise an error instead of unexpectedly crashing
+        primitives_supported = ["int", "bool", "float"]
```
NaderAlAwar, line 161: Will this not work with numpy types then?

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +124,165 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+        entity_tree: ast.AST = None
+        
+        #*2 Changing annotations for the needed workunit definitions
+        for node in self.tree.body:
+
+            # Check for the workunit we need
+            if check_entity(node, self.pk_import) and updated_types.workunit.__name__ == node.name:
```
NaderAlAwar, line 168: This condition is kind of hard to understand. Can you move this logic to a separate method?

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +124,165 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+                
+                entity_tree = node
+
+                # if modifications to layout decorator is needed
+                if len(updated_types.layout_change):
+                    node.decorator_list = self.fix_viewlayout(node, updated_types.layout_change)
+
+                for arg_obj in node.args.args:
+                    if arg_obj.arg in updated_types.inferred_types:
```
NaderAlAwar, line 177: Again it's not clear to me what the condition here is checking for. Maybe add a method to the `updated_types` class?

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +124,165 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+                        if update_type in primitives_supported:
+                            arg_obj.annotation = ast.Name(id=update_type, ctx=ast.Load())
+                                                        
+                        elif "View" in update_type:
+                            # update_type = View1D:double
+                            view_type, dtype = update_type.split(':')
+
+                            arg_obj.annotation = ast.Subscript(
+                                value = ast.Attribute(
+                                    value = ast.Name(id="pk", ctx=ast.Load()),
```
NaderAlAwar, line 189: Instead of "pk", you should be using `pk_import`. That way if a user has `import pykokkos as pyk' it would still work.

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +124,165 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+                        
+                        elif "pk.TeamMember" in update_type:
+
+                            arg_obj.annotation = ast.Attribute(
+                                value = ast.Name(id = "pk", ctx = ast.Load()),
+                                attr = "TeamMember",
+                                ctx = ast.Load()
+                            )
+                        else:
+                            raise ValueError("ERROR: Unsupported type inference")
```
NaderAlAwar, line 223: Maybe print the type as well here

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +124,165 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+                        
+                break
```
NaderAlAwar, line 225: Why do we need a break statement here?

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +124,165 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+
+        return entity_tree
+
+
+    def fix_viewlayout(self, node : ast.AST, layout_change: Dict[str, str]):
```
NaderAlAwar, line 230: For consistency, rename this to `fix_view_layout`

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +124,165 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+        '''
+        node: ast object for the workunit
+        layout_change: Dict that stores view variable identifier as keys against the layout type
+
+        This function returns the modified workunit ast with corrent layout decorators
+        '''
```
NaderAlAwar, line 236: Same as before, look at the format we use for docstrings

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +124,165 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+        if len(node.decorator_list) and isinstance(node.decorator_list[0], ast.Call):
+            call_obj = node.decorator_list[0]
+            for keyword_obj in call_obj.keywords:
+                if keyword_obj.arg in layout_change:
+                    # user provided do not modify
+                    del layout_change[keyword_obj.arg]
```
NaderAlAwar, line 242: This block of code is hard to understand. Can you add some comments?

pykokkos/core/parsers/parser.py
NaderAlAwar: You are also missing the return type

pykokkos/core/parsers/parser.py
```diff
@@ -154,7 +301,7 @@ def is_classtype(node: ast.stmt, pk_import: str) -> bool:
                 if (attribute.value.id == pk_import
                         and Decorator.is_kokkos_classtype(attribute.attr)):
                     return True
-
+
```
NaderAlAwar, line 304: This change is not needed

pykokkos/core/visitors/workunit_transformer.py
```diff
@@ -0,0 +1,18 @@
+from _ast import FunctionDef
+from ast import NodeTransformer
+from typing import Any
+
+class RemoveTransformer(NodeTransformer):
```
NaderAlAwar, line 5: Please add a docstring that explains the purpose of this class. Look at the docstrings for other classes that have them

pykokkos/core/visitors/workunit_visitor.py
```diff
@@ -33,7 +33,7 @@ def visit_FunctionDef(self, node: ast.FunctionDef) -> Union[str, Tuple[str, cppa
         else:
             operation: Optional[str] = self.get_operation_type(node)
             if operation is None:
-                self.error(node.args, "Incorrect types in workunit definition")
+                self.error(node.args, "Incorrect types in workunit definition: " + node.name)
```
NaderAlAwar, line 36: use an f-string instead of  `+ node.name`

examples/kokkos-tutorials/functor/01_n.py
NaderAlAwar, line 1: This file is not needed and can be deleted

NaderAlAwar (changes requested):
This is very good work, thanks Hannan! A few things that need to be fixed:
1. In a lot of places, you added newlines and whitespace in places where they are not needed. Please go over the diff in the PR and remove those. In particular, you added two consecutive newlines in quite a few places, please remove those.
2. Some of the if statements you added are hard to understand. What I usually do is I move the logic of the condition to a method and call that in the if statement. For example, see [here](https://github.com/kokkos/pykokkos/blob/1514b5f8e16df26d5376094d84b48cc63c82c5d0/pykokkos/core/compiler.py#L262)
3. Please add a test that uses standalone with no types
4. Various other comments I left

Additionally, for the examples, instead of adding a new `notype` directory to standalone, I would like you to replace the old standalone examples with the new syntax that does not require static types. Also get rid of the `_n` suffix in the name of these examples.

pykokkos/core/compiler.py
HannanNaeem: When there are no types to be inferred; the user already gave the annotations. I can either add a comment here to point this out or add it to the doc string, ... or both

pykokkos/core/compiler.py
HannanNaeem: :heavy_check_mark:

pykokkos/core/visitors/workunit_transformer.py
HannanNaeem: Deleted this, as this was only needed for functors

pykokkos/core/visitors/workunit_visitor.py
HannanNaeem: Reverting this change, this was just to see where my implementation was breaking.

pykokkos/core/parsers/parser.py
HannanNaeem: Eliminated for loop altogether; it was coming from functor implementation

pykokkos/core/parsers/parser.py
HannanNaeem: Added `Type already provided by the user`. Unfortunately the AST is always convoluted like this, a method will again contain the same statement

pykokkos/core/parsers/parser.py
HannanNaeem: Moved this to a method `filter_layout_change` so it can be better documented and read

pykokkos/core/parsers/parser.py
NaderAlAwar: You're still missing `:param` . For example, here it would be `:param entity: ...`. You're also missing the colon just before returns, so it should be `:returns:`. See the docstring [here](https://github.com/kokkos/pykokkos/blob/6b894e88b360fcaf21c152b739172ed0cc2ad013/pykokkos/core/compiler.py#L64).

pykokkos/core/parsers/parser.py
NaderAlAwar: I'm assuming this is fixed now? If so you can resolve it

pykokkos/core/parsers/parser.py
```diff
@@ -121,21 +121,181 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+                )
+
+            return [call_obj]
+
+        # no change needed
+        return node.decorator_list
+
+    @staticmethod
+    def filter_layout_change(node: ast.AST, working_dict: Dict[str, str]) -> Dict[str, str]:
+        #MARK ADD DOCSTRING
```
NaderAlAwar: Don't forget to add this

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,234 @@
...
+    handled_args: Processed arguments passed to the dispatch
+    args: raw arguments passed to the dispatch
+    passed_kwargs: raw keyword arguments passed to the dispatch
+
+    returns: UpdateTypes object or None if there are no annotations to be inferred
+
+    This function will infer first, the datatypes of policy arguments, and then any additional arguments
+    For readability: params/parameters will refer to the workunit signature
+                    args/argument will refer to the actual raw values passed during execution
+    '''
```
NaderAlAwar, line 110: Fix this docstring as well

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,234 @@
...
+        policy_params += 1
+    # accumulator + lass_pass
+    if parallel_type == "parallel_scan":
+        policy_params += 2
+
+    # Handling policy parameters
+    for i in range(policy_params):
+        param = param_list[i]
+
+        if param.annotation is inspect._empty:
```
NaderAlAwar, line 128: Replace this with

```python
if param.annotation is not inspect._empty:
    continue
```

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,234 @@
...
+                total_dims = len(handled_args.policy.begin) 
+                if i < total_dims:
+                    updated_types.inferred_types[param.name] = "int"
+                    updated_types.is_arg.add(param.name)
+            else:
+                raise ValueError("Automatic annotations not supported for this policy")
+
+            # last policy param for parallel reduce and second last for parallel_scan is always the accumulator; the default type is double
+            if i == policy_params - 1 and parallel_type == "parallel_reduce" or i == policy_params - 2 and parallel_type == "parallel_scan":
+                updated_types.inferred_types[param.name] = "Acc:double"
```
NaderAlAwar, line 151: Above you use `pk.TeamMember`, but here you just use `Acc`, you should be more consistent. I think you can just remove `pk.` from `TeamMember` above right?

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,234 @@
...
+    else:
+        raise ValueError(f"ERROR: incorrect number of arguments {len(unpacked)}")
+
+    if isinstance(policy, int):
+        policy = RangePolicy(km.get_default_space(), 0, policy)
+
+    return HandledArgs(name, policy, workunit, view, initial_value)
+
+
+def get_annotations(parallel_type: str, handled_args: HandledArgs, *args, passed_kwargs) -> UpdatedTypes:
```
NaderAlAwar, line 98: If this is meant to return `None` in some cases, the return type should be `Optional[UpdatedTypes]`

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,234 @@
...
+            if i == policy_params - 1 and parallel_type == "parallel_scan":
+                updated_types.inferred_types[param.name] = "bool"
+                updated_types.is_arg.add(param.name)
+
+    if len(param_list) == policy_params:
+        if not len(updated_types.inferred_types): return None
+        return updated_types
+
+    # Handle Keyword args, make sure they are treated by queuing them in args
+    if len(passed_kwargs.keys()):
```
NaderAlAwar, line 163: you don't need the `.keys()` here

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,234 @@
...
+        # add value to arguments list so the value can be assessed
+        for param in param_list[policy_params:]:
+            if param.name in passed_kwargs:
+                args_list.append(passed_kwargs[param.name])
+
+    # Handling arguments other than policy args, they begin at value_idx in args list
+    value_idx: int = 3 if handled_args.name != None else 2
```
NaderAlAwar, line 170: It's not clear to me why this is 3 or 2

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,234 @@
...
+    supported_np_dtypes = [attr for attr in dir(DataType) if not attr.startswith("__")]
+
+    for i in range(policy_params , len(param_list)):
+        param = param_list[i]
+        value = args_list[value_idx + i - policy_params]
+
+        if isinstance(value, View) and value.layout != Layout.LayoutDefault:
+            updated_types.layout_change[param.name] = "LayoutRight" if value.layout == Layout.LayoutRight else "LayoutLeft"
+
+        if param.annotation is not inspect._empty:
```
NaderAlAwar, line 187: This should probably come before the previous if statement right?

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,234 @@
...
+                raise TypeError("Cannot infer datatype for view:", param.name)
+
+            param_type = "View"+str(len(value.shape))+"D:"+view_dtype
+
+        updated_types.inferred_types[param.name] = param_type 
+        updated_types.is_arg.add(param.name)
+
+    if not len(updated_types.inferred_types): return None
+
+    return updated_types
```
NaderAlAwar, line 218: This function is too long, it should probably be split up into multiple functions

NaderAlAwar (changes requested):
See comments

pykokkos/core/parsers/parser.py
```diff
@@ -68,10 +68,10 @@ def get_import(self) -> str:
...
 
         package: str = "pykokkos"
-        for node in self.tree.body:
-            if isinstance(node, ast.Import):
-                if node.names[0].name == package:
-                    alias: ast.alias = node.names[0]
+        for entity_tree in self.tree.body:
+            if isinstance(entity_tree, ast.Import):
+                if entity_tree.names[0].name == package:
+                    alias: ast.alias = entity_tree.names[0]
```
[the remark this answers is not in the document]
HannanNaeem: Indeed, a product of taking a keyboard shortcut I didnt understand for a test drive xDD

pykokkos/core/parsers/parser.py
HannanNaeem: Fixed in the new commit

pykokkos/interface/type_inference.py
HannanNaeem: True, I have removed the pk prefix

pykokkos/interface/type_inference.py
HannanNaeem: updating comments to:
`  # Handling arguments other than policy args, they begin at value_idx in args list
   # e.g idx=3 -> parallel_for("label", policy, workunit, <other args>...) or if "label" is missing: 2 `

pykokkos/interface/type_inference.py
HannanNaeem: No because we allow mix and match. If a view has annotations but is missing layout decorator, the decorator won't be corrected. Hence we check for layout first regardless, and then move to checking annotations

pykokkos/interface/type_inference.py
HannanNaeem: Please resolve this if you think this should be good

pykokkos/interface/type_inference.py
HannanNaeem: Please resolve this if you think the explanation suffices

examples/BabelStream/standalone/babel_stream.py
```diff
@@ -12,23 +12,23 @@ def init_arrays(index: int, a: pk.View1D[pk.double], b: pk.View1D[pk.double], c:
     c[index] = initC
 
 @pk.workunit
-def copy(index: int, a: pk.View1D[pk.double], c: pk.View1D[pk.double]):
+def copy(index, a, c):
```
gliga, line 15: Going Pythonic is nice, but now that we loose types, we should update the variable names to reflect something behind data being in those variables. If we do not want to use views any more, do we talk about arrays or lists?

gliga: same comment for other functions/files

examples/ParRes/standalone/nstream.py
```diff
@@ -4,15 +4,15 @@
 import sys
 
 @pk.workunit
-def init_view(i: int, inp: pk.View1D[pk.double], init: float):
+def init_view(i, inp, init):
     inp[i] = init
 
 @pk.workunit
-def nstream( i: int, A: pk.View1D[pk.double], B: pk.View1D[pk.double], C: pk.View1D[pk.double], scalar: float):
+def nstream( i, A, B, C, scalar):
```
gliga, line 11: in some files we use lower case letters for arguments and in some cases uppercase.  Let's use lowercase always until we have already convention otherwise (in which case we should update those that are lowercase to uppercase).

pykokkos/core/parsers/parser.py
```diff
@@ -40,13 +40,13 @@ def __init__(self, path: str):
...
         :param path: the path to the file
         """
-
         self.lines: List[str]
         self.tree: ast.Module
         with open(path, "r") as f:
             self.lines = f.readlines()
             self.tree = ast.parse("".join(self.lines))
 
+
```
gliga, line 49: why are we doing formatting

pykokkos/core/runtime.py
```diff
@@ -191,6 +193,7 @@ def import_module(self, module_name: str, module_path: str):
 
         spec = importlib.util.spec_from_file_location(module_name, module_path)
         module = importlib.util.module_from_spec(spec)
+
```
gliga, line 196: if not necessary, I suggest we do formatting in a separate pass

examples/kokkos-benchmarks/standalone/notype/bytes_and_flops.py
```diff
@@ -0,0 +1,108 @@
+import argparse
```
gliga, line 1: is this a new example? let's keep this for the next PR and we focus this PR on removing types.

examples/kokkos-benchmarks/standalone/notype/gather.py
```diff
@@ -0,0 +1,105 @@
+import argparse
```
gliga, line 1: same as above

examples/kokkos-benchmarks/standalone/notype/gups.py
```diff
@@ -0,0 +1,86 @@
+import argparse
+import random
```
gliga, line 2: same comment as earlier: keep new examples for the next PR, so we can review them independently

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +123,186 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
                 entity = PyKokkosEntity(style, cppast.DeclRefExpr(name), node, (self.lines[start:stop], start), self.path, self.pk_import)
                 entities[name] = entity
 
         return entities
 
+
+    def fix_types(self, entity: PyKokkosEntity, updated_types: UpdatedTypes) -> ast.AST:
+        '''
+        Inject (into the entity AST,) the missing annotations for datatypes that have been inferred
```
gliga, line 143: AST, -> AST

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +123,186 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
         return entities
 
+
+    def fix_types(self, entity: PyKokkosEntity, updated_types: UpdatedTypes) -> ast.AST:
+        '''
+        Inject (into the entity AST,) the missing annotations for datatypes that have been inferred 
+
+        entity: Pykokkos entity whose AST will be patched - the entity being compiled/translated
+        updated_types: UpdatedTypes object that contains info about inferred types for this entity
+        returns: the updated entity AST after injecting correct annotations (from updated_types) for datatypes
```
gliga, line 147: "." at the end

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +123,186 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+        '''
+
+        style: PyKokkosStyles = entity.style
+        assert style is PyKokkosStyles.workunit and updated_types is not None
+
+        # For now, just so we can raise an error instead of unexpectedly crashing
+        primitives_supported = ["int", "bool", "float"]
```
gliga, line 154: extract into global; we do not need each time a new list

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +123,186 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+
+        entity_tree: ast.AST = entity.AST
+
+        # if modifications to layout decorator is needed
+        if len(updated_types.layout_change):
+            entity_tree.decorator_list = self.fix_view_layout(entity_tree, updated_types.layout_change)
+
+        for arg_obj in entity_tree.args.args:
+            # Type already provided by the user
+            if arg_obj.arg not in updated_types.inferred_types:
```
gliga, line 164: you could check here if the user type matches inferred type

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +123,186 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+                call_obj = node.decorator_list[0]
+            else: 
+                call_obj= ast.Call()
+                call_obj.func = ast.Attribute(value=ast.Name(id=self.pk_import, ctx=ast.Load()), attr='workunit', ctx=ast.Load())
+                call_obj.args = []
+                call_obj.keywords = []
+
+            for view, layout in layout_change.items():
+                call_obj.keywords.append(
+                    ast.keyword(
```
gliga, line 259: can this be extracted into a function/method

pykokkos/core/runtime.py
```diff
@@ -111,7 +113,7 @@ def run_workunit(
                 raise RuntimeError("ERROR: operation cannot be None for Debug")
             return run_workunit_debug(policy, workunit, operation, initial_value, **kwargs)
 
-        members: Optional[PyKokkosMembers] = self.precompile_workunit(workunit,policy.space)
+        members: Optional[PyKokkosMembers] = self.precompile_workunit(workunit,policy.space,updated_types)
```
gliga, line 116: missing space after ","

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,279 @@
+import inspect
+from dataclasses import dataclass
```
gliga, line 2: should this file be args_type_inference?

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,279 @@
...
+@dataclass
+class UpdatedTypes:
+    """
+    Class for storing inferred type annotation information 
+    (Making Pykokkos more pythonic by automatically inferring types)
+    """
+
+    workunit: Callable
+    inferred_types: Dict[str, str] # type information stored as string: identifier -> type
+    is_arg: set[str]
```
gliga, line 32: should this be boolean if starts with `is`

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,279 @@
...
+    layout_change: Dict[str, str] # if layout for a view is not Layout.Default it will be stored here
+
+
+def handle_args(is_for: bool, *args) -> HandledArgs:
+    """
+    Handle the *args passed to parallel_* functions
```
gliga, line 38: Handle the *args passed to parallel_* functions =>
Handle the *args passed to functions

at this point it is not relevant any longer it is from parallel_ function; could be any in theory

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,279 @@
...
+
+    # Handling policy parameters
+    updated_types = infer_policy_args(param_list, policy_params, handled_args.policy, parallel_type, updated_types)
+
+    # Policy parameters are the only parameters
+    if len(param_list) == policy_params:
+        if not len(updated_types.inferred_types): return None
+        return updated_types
+
+    # Handle Keyword args, make sure they are treated by queuing them in args
```
gliga, line 129: Keyword args => keyword args

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,279 @@
...
+            param_type = "View"+str(len(value.shape))+"D:"+view_dtype
+
+        updated_types.inferred_types[param.name] = param_type 
+        updated_types.is_arg.add(param.name)
+
+    return updated_types
+
+
+def get_pk_datatype(value):
+    '''
```
gliga, line 267: what if value is `None`

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,279 @@
...
+
+    dtype = None
+    if isinstance(value, DataType):
+        dtype = str(value.name)
+
+    elif inspect.isclass(value) and issubclass(value, DataTypeClass):
+        dtype = str(value.__name__)
+
+    return dtype
```
gliga, line 279: should we raise an exception or return `None` if no match

pykokkos/interface/type_inference.py
```diff
@@ -0,0 +1,279 @@
...
+
+        if isinstance(value, View) and value.layout != Layout.LayoutDefault:
+            updated_types.layout_change[param.name] = "LayoutRight" if value.layout == Layout.LayoutRight else "LayoutLeft"
+
+        if param.annotation is not inspect._empty:
+            continue
+
+        param_type = type(value).__name__
+
+        # switch integer values over 31 bits (signed positive value) to pk.int64
```
gliga, line 239: we say pk.int64 but below we use numpy:int64

gliga (commented):
Nice overall, some comments.

pykokkos/interface/type_inference.py
HannanNaeem: It is raised after this function returns to the the caller

examples/kokkos-benchmarks/standalone/notype/gups.py
HannanNaeem: These were no-types versions of existing examples. As discussed over slack I am replaced the original examples themselves now. Resolving this.

examples/kokkos-benchmarks/standalone/notype/gather.py
```diff
@@ -0,0 +1,105 @@
...
+import random
+from typing import Tuple
+
+import pykokkos as pk
+
+
+# use double type and unroll=8
+@pk.workunit
+def benchmark(i, K, F, A, B, C, connectivity):
```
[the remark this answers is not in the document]
HannanNaeem: Nader pointed out that other formats for the same examples have the data types annotated. Never the less, I still think this is a concern. Will discuss with you guys

pykokkos/core/parsers/parser.py
HannanNaeem: We don't infer any type the user already provided for. We decided this to let user "force" a type in cases were inference yields unwanted results

pykokkos/core/parsers/parser.py
```diff
@@ -123,19 +123,186 @@ def get_entities(self, style: PyKokkosStyles) -> Dict[str, PyKokkosEntity]:
...
+
+    def fix_view_layout(self, node : ast.AST, layout_change: Dict[str, str]) -> List[ast.Call]:
+        '''
+        Add (to the node AST) the missing layout decorators for pykokkos views
+
+        node: ast object for the entity
+        layout_change: Dict that maps [view -> layout]
+        '''
+
+        assert len(node.decorator_list), "Decorator cannot be missing for pykokkos workunit"
```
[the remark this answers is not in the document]
HannanNaeem: Added.

pykokkos/core/parsers/parser.py
HannanNaeem: Done

pykokkos/interface/type_inference.py
HannanNaeem: I think I will refactor this once I know how to deal with workunit bodies. It might be that we can expand/use the same file

pykokkos/interface/type_inference.py
HannanNaeem: Got rid of this altogether

pykokkos/interface/type_inference.py
HannanNaeem: This method is being moved to type_inference from parallel_dispatch so we have everything argument related in one place. (not authored by me) But it does strictly deal with only parallel_<> functions. Which is why the specification is necessary

I think we should leave it as it is until we remove parallel_for etc.

pykokkos/interface/type_inference.py
HannanNaeem: This is not possible. I have updated the argument name and the doc string to reflect why. This function should only be called on view.dtype (which cannot be none)

pykokkos/interface/type_inference.py
HannanNaeem: fixed

examples/BabelStream/standalone/babel_stream.py
NaderAlAwar: In this case, I think these variable names are copied from the original C++/Kokkos examples, so we can keep it for now.

examples/ParRes/standalone/nstream.py
NaderAlAwar: Same as above, I believe this is copied from the original examples. It's fine to leave it IMO.

NaderAlAwar (approved):
Looks good to me

examples/kokkos-benchmarks/standalone/notype/gather.py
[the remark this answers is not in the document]
HannanNaeem: Done. The ones not modified have detailed comments at the call to the workunit

examples/BabelStream/standalone/babel_stream.py
HannanNaeem: Changed anyways

PR #209, opened 2023-10-25
Add support for execution spaces instances

NaderAlAwar:


examples/pykokkos/streams.py
```diff
@@ -0,0 +1,27 @@
...
+    elif x == 2:
+        pk.printf("Stream 2\n")
+
+if __name__ == "__main__":
+    space = pk.Cuda
+
+    s1 = cp.cuda.Stream()
+    s2 = cp.cuda.Stream()
+
+    instance1 = pk.ExecutionSpaceInstance(space, s1)
```
gliga, line 20: ExecutionSpaceInstance => ExecutionSpace

examples/pykokkos/streams.py
```diff
@@ -0,0 +1,27 @@
...
+        pk.printf("Default stream\n")
+    elif x == 1:
+        pk.printf("Stream 1\n")
+    elif x == 2:
+        pk.printf("Stream 2\n")
+
+if __name__ == "__main__":
+    space = pk.Cuda
+
+    s1 = cp.cuda.Stream()
```
gliga, line 17: Should we use pk.Stream to avoid importing cupy

examples/pykokkos/streams.py
gliga: if we are instantiating a stream on top of a space, we could also go with `space.make(stream)`. (And potentially have also a version that does not accept a stream but uses a new stream for each call.)

pykokkos/core/translators/bindings.py
```diff
@@ -311,7 +312,7 @@ def generate_call(operation: str, functor: str, members: PyKokkosMembers, tag: c
         call += f"}} else {{ {custom_call} }}"
 
     call += generate_copy_back(members)
-    call += generate_fence_call()
+    # call += generate_fence_call()
```
gliga, line 315: including a comment to explain the reason this is commented out would be good

pykokkos/interface/execution_space.py
```diff
@@ -22,4 +22,38 @@ def is_host_execution_space(space: ExecutionSpace) -> bool:
     if space is ExecutionSpace.Default:
         space = km.get_default_space()
 
-    return space in {ExecutionSpace.OpenMP, ExecutionSpace.Pthreads, ExecutionSpace.Serial}
\ No newline at end of file
+    return space in {ExecutionSpace.OpenMP, ExecutionSpace.Threads, ExecutionSpace.Serial}
+
+class ExecutionSpaceInstance:
+    """
+    An instance of the execution space, corresponding to Cuda/HIP
```
gliga, line 29: If this only corresponds to streams, should we call it that way?

gliga (commented):
Looks good.

A few comments for a future PR.
The difference between execution space and execution space instance is confusing: it could be due to names used, so considering switching to Stream for instances.

examples/pykokkos/streams.py
NaderAlAwar: The motivation here is that users might have pre-existing CuPy code they want to integrate with pykokkos, and this example shows how.

examples/pykokkos/streams.py
NaderAlAwar: That does seem like a better interface. I'll address this in a new PR

pykokkos/interface/execution_space.py
NaderAlAwar: Technically users can create an instance for OpenMP (which doesn't have streams), although I don't think it does anything right now. Future execution spaces might have use for this and they might not use the terminology "Streams".

examples/pykokkos/streams.py
gliga: but the example only uses cupy to create streams not to use cupy arrays; would be good to then use cupy for all parts

pykokkos/interface/execution_space.py
gliga: Stream would at least reflect the current usage.  In the future, you could reconsider the name if needed.  `ExecutionSpaceInstace` sounds like an instance of a class, but then instance of an instance sounds like a meta level. This is something to be discussed in the next pykokkos meeting, but seems unnecessarily complex in my view.

PR #214, opened 2023-11-21
Improve performance of kernel fusion by eliminating unnecessary deep copy

NaderAlAwar:
Remove `deepcopy()` from the parser when retrieving workunits to avoid unnecessary overhead. Previously it was done for every workunit AST since fusion requires that we modify the ASTs. We still `deepcopy()` but only for fused workunits and only after we check whether the fused workunit has already been compiled, greatly reducing the overhead by PyKokkos.

### DIFF
--- a/pykokkos/core/visitors/workunit_visitor.py
+++ b/pykokkos/core/visitors/workunit_visitor.py
@@ -63,6 +63,10 @@ class WorkunitVisitor(PyKokkosVisitor):
         args: List[ast.arg] = node.args.args
         last_arg: ast.arg = args[0]
 
+        # Find the last argument in the workunit function definition that is not
+        # a view or a field. This is important as this argument could be the thread ID,
+        # the accumulator, or a boolean, which would help determine what the operation
+        # is (for, reduce, or scan)
         for arg in args:
             arg_name = cppast.DeclRefExpr(arg.arg)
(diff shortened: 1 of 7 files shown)


pykokkos/interface/parallel_dispatch.py
```diff
@@ -61,9 +61,9 @@ def parallel_for(*args, **kwargs) -> None:
         handled_args.name,
         handled_args.policy,
         handled_args.workunit,
+        "for",
```
gliga, line 64: minor: do we want enum for operations (rather than strings)

pykokkos/core/visitors/workunit_visitor.py
```diff
@@ -61,7 +61,14 @@ def get_operation_type(self, node: ast.FunctionDef) -> Optional[str]:
         """
 
         args: List[ast.arg] = node.args.args
-        last_arg: ast.arg = args[-1]
+        last_arg: ast.arg = args[0]
+
+        for arg in args:
```
gliga, line 66: comment on this loop would be valuable

gliga (approved):
LG

pykokkos/interface/parallel_dispatch.py
NaderAlAwar: Yeah this should be done, probably in a later PR

Issue #217, opened 2023-11-30
Several examples need minor sanitization to avoid runtime error at Kokkos::finalize


pkestene:
Just to mention that several examples fail because kokkos parallel lauch is done directly in python `__main__`,  see e.g. : [babel_stream.py#L46](https://github.com/kokkos/pykokkos/blob/main/examples/BabelStream/functor/babel_stream.py#L46)

```shell
> python babel_stream.py
Validation failed on a[]. Average error 5.148412890317342e-08
Validation failed on b[]. Average error 2.1451720376358397e-08
Validation failed on c[]. Average error 7.50810213173628e-08
Function    MBytes/sec  Min (sec)   Max         Average     
Copy        819.226     0.00064     0.00104     0.00071     
Mul         836.525     0.00063     0.00106     0.00071     
Add         1240.474    0.00063     0.00097     0.00071     
Triad       1255.108    0.00063     0.00158     0.00072     
Dot         900.869     0.00058     0.00097     0.00066     
terminate called after throwing an instance of 'std::runtime_error'
  what():  Kokkos allocation "" is being deallocated after Kokkos::finalize was called

Abandon (core dumped)
```

We just need to introduce a separate python function, and call it from `__main__` to avoid this error

PR #218, opened 2023-11-30
Fix examples that fail because Kokkos::finalize called too early

pkestene:
fixing #217


NaderAlAwar (approved):
Looks good, thanks for taking care of this

PR #226, opened 2023-12-18
Removing dependence on inspect

HannanNaeem:
This PR removes dependence on inspect library for type inference pipeline. This restructure includes:
- Reading of types now happens in the runtime object in `runtime.py` using the parser.
- This means that the parser is now initialized first in `runtime.py` using the compiler object already available
- Original parameters and their annotation are now read from the workunit AST itself.
- Runtime object now has a cache for the original state of the arguments node in the AST (to reset the arguments for the next call to the workunit)

### DIFF
--- a/pykokkos/core/runtime.py
+++ b/pykokkos/core/runtime.py
@@ -10,7 +10,10 @@ from pykokkos.core.fusion import fuse_workunit_kwargs_and_params
 from pykokkos.core.keywords import Keywords
 from pykokkos.core.translators import PyKokkosMembers
 from pykokkos.core.visitors import visitors_util
-from pykokkos.core.type_inference import UpdatedTypes, UpdatedDecorator, get_types_signature
+from pykokkos.core.type_inference import (
+    UpdatedTypes, UpdatedDecorator, 
+    get_types_signature, get_annotations, get_views_decorator
+)
 from pykokkos.interface import (
     DataType, ExecutionPolicy, ExecutionSpace, MemorySpace,
     RandomPool, RangePolicy, TeamPolicy, View, ViewType,
@@ -19,7 +22,7 @@ from pykokkos.interface import (
 import pykokkos.kokkos_manager as km
 
 from .compiler import Compiler
-from .module_setup import ModuleSetup
+from .module_setup import ModuleSetup, get_metadata
 from .run_debug import run_workload_debug, run_workunit_debug
 
 
@@ -103,8 +106,6 @@ class Runtime:
         policy: ExecutionPolicy,
         workunit: Union[Callable[..., None], List[Callable[..., None]]],
         operation: str,
-        updated_decorator: UpdatedDecorator,
-        updated_types: Optional[UpdatedTypes] = None,
         initial_value: Union[float, int] = 0,
         **kwargs
     ) -> Optional[Union[float, int]]:
@@ -128,8 +129,21 @@ class Runtime:
             if operation is None:
                 raise RuntimeError("ERROR: operation cannot be None for Debug")
             return run_workunit_debug(policy, workunit, operation, initial_value, **kwargs)
-
+        '''
+        TODO
+        1- Initialize parser for the self.compiler
+        2- Get updated types and decorator list
+        3- Construct types signature 
+        4- Make the module setup
+        5- let it rip
+        '''
+        metadata = get_metadata(workunit)
+        parser = self.compiler.get_parser(metadata.path)
+        entity_AST = parser.get_entity(metadata.name).AST
+        updated_types: UpdatedTypes = get_annotations(f"parallel_{operation}", entity_AST, workunit, policy, passed_kwargs=kwargs)
+        updated_decorator: UpdatedDecorator = get_views_decorator(workunit, passed_kwargs=kwargs)
         types_signature: str = get_types_signature(updated_types, updated_decorator, execution_space)
+
         members: Optional[PyKokkosMembers] = self.precompile_workunit(workunit, execution_space, updated_decorator, updated_types, types_signature, **kwargs)
         if members is None:
             raise RuntimeError("ERROR: members cannot be none")
--- a/pykokkos/core/runtime.py
+++ b/pykokkos/core/runtime.py
@@ -3,6 +3,7 @@ from pathlib import Path
 import sys
 from typing import Any, Callable, Dict, Optional, Tuple, Type, Union, List
 import sysconfig
+import ast
 
 import numpy as np
 
@@ -22,7 +23,7 @@ from pykokkos.interface import (
 import pykokkos.kokkos_manager as km
 
 from .compiler import Compiler
-from .module_setup import ModuleSetup, get_metadata
+from .module_setup import ModuleSetup, EntityMetadata, get_metadata
 from .run_debug import run_workload_debug, run_workunit_debug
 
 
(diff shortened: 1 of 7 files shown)


pykokkos/core/runtime.py
```diff
@@ -129,13 +132,69 @@ def run_workunit(
...
+        updated_types: UpdatedTypes = None
+        updated_decorator: UpdatedDecorator = None
+        types_signature: str = None
+
+        if is_standalone_workunit:
+            if is_missing_annotations:
+                updated_types = get_annotations(f"parallel_{operation}", workunit_tree_tup, policy, passed_kwargs=kwargs)
+            updated_decorator = get_views_decorator(workunit_tree_tup, passed_kwargs=kwargs)
+            types_signature = get_types_signature(updated_types, updated_decorator, execution_space)
+
```
NaderAlAwar, line 191: You should move everything you added to `run_workunit` to `args_type_inference.py`. This is too much detail for the runtime, it should be replaces by a single method call

pykokkos/core/runtime.py
```diff
@@ -180,7 +241,7 @@ def execute(
 
         module = self.import_module(module_setup.name, module_path)
 
-        args: Dict[str, Any] = self.get_arguments(entity, members, space, policy, **kwargs)
+        args: Dict[str, Any] = self.get_arguments(entity, members, space, policy, entity_trees=entity_trees, **kwargs)
```
NaderAlAwar, line 244: You don't need `entity_trees=entity_trees`, just pass in `entity_trees`

pykokkos/core/type_inference/args_type_inference.py
```diff
@@ -112,39 +113,55 @@ def handle_args(is_for: bool, *args) -> HandledArgs:
 
     return HandledArgs(name, policy, workunit, view, initial_value)
 
+def check_missing_annotations(param_list: List[ast.arg]) -> bool:
+    '''
+    :param param_list: list of ast.arg objects to run annotation check against
```
NaderAlAwar, line 118: Add a short description as well

pykokkos/core/type_inference/args_type_inference.py
```diff
@@ -112,39 +113,55 @@ def handle_args(is_for: bool, *args) -> HandledArgs:
...
+
+    missing = False
+    for param in param_list:
+        if param.annotation is None:
+            missing = True
+            break
+    return missing
 
-def get_annotations(parallel_type: str, handled_args: HandledArgs, *args, passed_kwargs) -> Optional[UpdatedTypes]:
+def get_annotations(parallel_type: str, workunit_tree_tuples: Union[Tuple[Callable, ast.AST], List[Tuple[Callable, ast.AST]]], policy: Union[ExecutionPolicy, int], passed_kwargs) -> Optional[UpdatedTypes]:
```
NaderAlAwar, line 130: we don't need the type in the name, just name it `workunit_trees`

pykokkos/core/type_inference/args_type_inference.py
```diff
@@ -172,7 +189,7 @@ def get_annotations(parallel_type: str, handled_args: HandledArgs, *args, passed
     return updated_types
 
 
-def get_views_decorator(handled_args: HandledArgs, passed_kwargs) -> UpdatedDecorator:
+def get_views_decorator(workunit_tree_tuples: List[Tuple[Callable, ast.AST]], passed_kwargs) -> UpdatedDecorator:
```
NaderAlAwar, line 192: same comments as above

pykokkos/interface/parallel_dispatch.py
```diff
@@ -54,16 +54,14 @@ def parallel_for(*args, **kwargs) -> None:
 
     handled_args: HandledArgs = handle_args(True, args)
 
-    updated_types: UpdatedTypes = get_annotations("parallel_for", handled_args, args, passed_kwargs=kwargs)
-    updated_decorator: UpdatedDecorator = get_views_decorator(handled_args, passed_kwargs=kwargs)
+    # updated_types: UpdatedTypes = get_annotations("parallel_for", handled_args, args, passed_kwargs=kwargs)
+    # updated_decorator: UpdatedDecorator = get_views_decorator(handled_args, passed_kwargs=kwargs)
```
NaderAlAwar, line 58: You can just delete them, no need to comment out

NaderAlAwar (changes requested):
Looks good so far. I might have more comments when you move the code from runtime

pykokkos/core/runtime.py
HannanNaeem: I tried doing that initially, but I realized we will need some design changes. The biggest one is where do we keep the cache for the original arg nodes. It made sense to me to keep them at `Runtime`, since the compiler and parser are accessibly by design here.

Now I can move this to `args_type_inference.py` but that will require, either

1.  Moving the cache there 
2. Interfacing with runtime object from `args_type_inference.py`, which in my opinion does not feel right (?) and will look convoluted.

I have however, abstracted out some parts of the details. Let me know if you think this is ok. Otherwise, I am happy to move the cache to `args_type_inference.py` as a global (or let me know if you see another elegant solution)

NaderAlAwar: If you're referring to `self.workunit_params` as the cache, I think it makes sense to move it to args_type_inference.py

pykokkos/core/runtime.py
```diff
@@ -3,14 +3,18 @@
 import sys
 from typing import Any, Callable, Dict, Optional, Tuple, Type, Union, List
 import sysconfig
-
+import ast
 import numpy as np
```
NaderAlAwar, line 7: Imports should be done in alphabetical order, so move `import ast` to the top. Also keep the newline between the standard library imports and the numpy import

pykokkos/core/runtime.py
```diff
@@ -34,6 +38,9 @@ def __init__(self):
         # cache module_setup objects using a workload/workunit and space tuple
         self.module_setups: Dict[Tuple, ModuleSetup] = {}
 
+        # original parameters for a workunit are preserved for inference changes in the future
+        self.workunit_params: Dict[str, ast.arguments] = {}
+
```
NaderAlAwar, line 43: I'm assuming this can be deleted now?

pykokkos/core/runtime.py
```diff
@@ -256,7 +291,7 @@ def get_arguments(
             else:
                 is_fused: bool = isinstance(entity, list)
                 if is_fused:
-                    kwargs, _ = fuse_workunit_kwargs_and_params(entity, kwargs)
+                    kwargs, _ = fuse_workunit_kwargs_and_params(entity_trees, kwargs)
```
NaderAlAwar, line 294: In order to simplify the changes made to `run_workunit`, I would get the list of ASTs here from the parser instead of passing them all the way down.

pykokkos/core/type_inference/args_type_inference.py
```diff
@@ -2,12 +2,14 @@
 from dataclasses import dataclass
 from typing import  Callable, Dict, Optional, Tuple, Union, List
 import hashlib
+import ast
+import pickle
```
NaderAlAwar, line 6: make sure the imports are in alphabetical order

pykokkos/core/type_inference/args_type_inference.py
```diff
@@ -50,6 +52,8 @@ class UpdatedDecorator:
 # DataType class has all supported pk datatypes, we ignore class members starting with __, add enum duplicate aliases
 SUPPORTED_NP_DTYPES = [attr for attr in dir(DataType) if not attr.startswith("__")] + ["float64", "float32"]
 
+# Cache for original argument nodes
+ORIGINAL_PARAMS: Dict[str, ast.arguments] = {}
```
NaderAlAwar, line 56: Leave a more detailed comment on what this is mapping from. What exactly does the `str` represent here?

NaderAlAwar (approved):
Looks good, thanks!

PR #230, opened 2023-12-25
doc: Include Docker image usage

gliga:
This PR describe the way our Docker image should be used for running and developing.

We should decide (separate from this PR) if we want to have `-dev` image independent from a default one that would be used for running examples.

(Once we have an image for CUDA as well, we should remove Installation instructions from the README file.)


NaderAlAwar (approved):
LGTM

PR #244, opened 2024-01-13
Fusion: add new 'naive' fusion strategy

NaderAlAwar:
Add a new 'naive' fusion strategy which fuses all consecutive `parallel_for` workunits. Reductions and scans are not fused for now. Also replace `PK_TRACE` environment variable with `PK_FUSION`, which can be set to "trace" for only delaying execution (but not actually fusing) or "naive" to delay execution and fuse using the previously mentioned strategy.

### DIFF
--- a/pykokkos/core/fusion/fuse.py
+++ b/pykokkos/core/fusion/fuse.py
@@ -1,7 +1,6 @@
 import ast
-import inspect
 import os
-from typing import Any, Callable, Dict, List, Set, Tuple, Union
+from typing import Any, Dict, List, Set, Tuple, Union
 
 def get_node_name(node: Union[ast.Attribute, ast.Name]) -> str:
     """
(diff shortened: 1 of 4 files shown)


gliga (approved):
This looks good.
I only did not get where do you ensure that local variables are unique (or maybe you do not need that because it is OK to reassign values as long as they are the same type, but what if they are not)?

NaderAlAwar:
> This looks good. I only did not get where do you ensure that local variables are unique (or maybe you do not need that because it is OK to reassign values as long as they are the same type, but what if they are not)?

I rename all local variables [here](https://github.com/kokkos/pykokkos/blob/0538c3cf7be5c59dcacdc4bf8d691c7e886fff85/pykokkos/core/fusion/fuse.py#L159). I added this in a previous PR.

PR #249, opened 2024-01-18
Fix Debug execution space crashing

NaderAlAwar:
This was caused by the recent introduction of execution space instances (which is not applicable to the Debug execution space). This PR fixes the crash by adding checks for Debug as necessary.


gliga (approved):
LG

PR #250, opened 2024-01-18
requirements.txt: Updated python version >=3.9, <3.12

HannanNaeem:
Explicitly defining the range of python versions supported -- this PR has a counterpart in pykokkos-base as well.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ pybind11>=2.11.1
 cupy>=12.2.0
 patchelf>=0.17.2
 pytest>=7.4.3
-python=3.11
+python>3.9,<3.12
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -2,4 +2,4 @@ pybind11>=2.11.1
 cupy>=12.2.0
 patchelf>=0.17.2
 pytest>=7.4.3
-python=3.11
+python>3.9,<3.12
```
NaderAlAwar, line 5: This is incorrect, should be >= 3.9

HannanNaeem: My bad, I thought I put 3.8

NaderAlAwar (approved):

PR #253, opened 2024-01-25
Implement loop fusion optimization inside kernels

NaderAlAwar:
This PR adds support for doing loop fusion inside a kernel. It can be enabled by setting the `PK_LOOP_FUSE` env variable. The conditions for two loops to be fused are as follows:

1. They must belong to the same scope OR their parent nodes are two for loops that can be fused.
2. They must have identical loop ranges. Right now this is checked by comparing the arguments of the two range calls as strings.
3. They must be adjacent OR the nodes between them are safe to move to before the first loop.
4. There must be no invalid dependencies (not implemented yet).


tests/test_loop_fusion.py
```diff
@@ -0,0 +1,164 @@
...
+
+    for j in range(1):
+        y: int = 4
+        x: int = y
+        # z: float = v[0]
+        # pk.printf("%d %d\n", y, j)
+        for k in range(2):
+            pk.printf("print 2 %d\n", k)
+
+def main():
```
gliga, line 155: can we format these as proper pytests so we can actually run them and have assertions in the end of each workunit?

gliga (approved):
Looks good except that comment about tests

PR #254, opened 2024-02-06
ufuncs: Added subtract support for 2D views

HannanNaeem:
`ufuncs.py`
  [+] Adding support for 2D views for the subtract util.
  [+] Added support for broadcasting views from scalar uptil 2D and anything inbetween (1D)

`pykokkos/__init__.py`
  [+] `broadcast_view(view_or_val, ontoThisView)` is an exposed utility

`test_ufuncs.py`
  [+] Added tests for subtract up til 2D -- these trigger broadcast too
  [-] Removed annotations in some test kernels to support scalars and views together

### DIFF
--- a/pre_compile_tools/pre_compile_ufuncs.py
+++ b/pre_compile_tools/pre_compile_ufuncs.py
@@ -55,6 +55,9 @@ def main():
                     func_obj(v, v)
                 except (NotImplementedError, KeyError):
                     pass
+                except ValueError:
+                    if "broadcast" in func[0]:
+                        pass
                 except TypeError:
                     try:
                         func_obj(v)
--- a/pre_compile_tools/pre_compile_ufuncs.py
+++ b/pre_compile_tools/pre_compile_ufuncs.py
@@ -27,7 +27,7 @@ def main():
     # level kernels/workunits directly
     filtered_function_list = []
     for f in function_list:
-        if not "impl" in f[0] and not f[0].startswith("_"):
+        if not "impl" in f[0] and not f[0].startswith("_") or ("broadcast_view" not in f[0]):
             filtered_function_list.append(f)
     # TODO: expand types and view dimensions for
     # ufunc pre-compilation as the support
@@ -55,9 +55,6 @@ def main():
                     func_obj(v, v)
                 except (NotImplementedError, KeyError):
                     pass
-                except ValueError:
-                    if "broadcast" in func[0]:
-                        pass
                 except TypeError:
                     try:
                         func_obj(v)
--- a/pre_compile_tools/pre_compile_ufuncs.py
+++ b/pre_compile_tools/pre_compile_ufuncs.py
@@ -27,7 +27,7 @@ def main():
     # level kernels/workunits directly
     filtered_function_list = []
     for f in function_list:
-        if not "impl" in f[0] and not f[0].startswith("_") or ("broadcast_view" not in f[0]):
+        if not "impl" in f[0] and not f[0].startswith("_") and "broadcast_view" not in f[0]:
             filtered_function_list.append(f)
     # TODO: expand types and view dimensions for
     # ufunc pre-compilation as the support
--- a/pykokkos/lib/ufuncs.py
+++ b/pykokkos/lib/ufuncs.py
@@ -1153,7 +1153,7 @@ def subtract(viewA, valB):
 
     if not is_scalar:
 
-        if not check_broadcastable_impl(viewA, valB):
+        if viewA.shape != valB.shape and not check_broadcastable_impl(viewA, valB): # if shape is not same check compatibility
             raise ValueError("Views must be broadcastable")
 
         # check if size is same otherwise broadcast and fix
--- a/pykokkos/lib/ufuncs.py
+++ b/pykokkos/lib/ufuncs.py
@@ -1059,7 +1059,7 @@ def broadcast_view(val, viewB):
         if not val.dtype == viewB.dtype: 
             raise ValueError("Broadcastable views must have same dtypes")
 
-    out = pk.View([dim for dim in viewB.shape], viewB.dtype)
+    out = pk.View(viewB.shape, viewB.dtype)
 
     if is_view:
         # if both 2D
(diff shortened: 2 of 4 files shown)


pykokkos/lib/ufuncs.py
```diff
@@ -959,17 +960,166 @@ def multiply(viewA, viewB):
...
+    out = pk.View([dim for dim in viewB.shape], viewB.dtype)
+
+    if is_view:
+        # if both 2D
+        if len(val.shape) == 2: #viewB must be 2 because of the val.shape < viewB.shape check
+            # figure which orientation is val (row or col)
+            col_wise = 1 if val.shape[1] == 1 else 0
+            inner_its = viewB.shape[0] if col_wise else viewB.shape[1]
+            outer_its = viewB.shape[1] if col_wise else viewB.shape[0]
+            policy = pk.TeamPolicy(outer_its, pk.AUTO)
```
NaderAlAwar, line 1071: Same comment as above, I would prefer a RangePolicy here

pykokkos/lib/ufuncs.py
```diff
@@ -959,17 +960,166 @@ def multiply(viewA, viewB):
...
+        if not is_view:
+            val = val[0] if len(val.shape) == 1 else val[0][0]
+
+    if is_view:
+        if not check_broadcastable_impl(val, viewB) or not val.shape < viewB.shape:
+            raise ValueError("Incompatible broadcast")
+        if not val.dtype == viewB.dtype: 
+            raise ValueError("Broadcastable views must have same dtypes")
+
+    out = pk.View([dim for dim in viewB.shape], viewB.dtype)
```
NaderAlAwar, line 1062: you should just be able to do `pk.View(viewB.shape, viewB.dtype)`

pykokkos/lib/ufuncs.py
```diff
@@ -959,17 +960,166 @@ def multiply(viewA, viewB):
...
+        else: # 1d to 2D
+            policy = pk.TeamPolicy(out.shape[0], pk.AUTO)
+            pk.parallel_for(policy, stretch_fill_impl_1d_into_2d, cols=viewB.shape[1], viewIn=val, viewOut=out)
+            
+        return out
+
+    # scalar
+
+    if len(viewB.shape) == 1:
+        out_1d = pk.View([viewB.shape[1] if len(viewB.shape)==2 else viewB.shape[0]])
```
NaderAlAwar, line 1082: Something is off with your logic here. The condition to enter this block is `len(viewB.shape) == 1`, but you are checking if it is equal to 2 when building the shape

pykokkos/lib/ufuncs.py
```diff
@@ -959,17 +960,166 @@ def multiply(viewA, viewB):
...
     out[tid] = viewA[tid] - viewB[tid]
 
+@pk.workunit
+def subtract_impl_2d(team_member, cols, viewA, viewB, viewOut):
+    j: int = team_member.league_rank()
+
+    def row_sub(i: int):
+        viewOut[j][i] = viewA[j][i] - viewB[j][i]
+    
+    pk.parallel_for(pk.TeamThreadRange(team_member, cols), row_sub)
```
NaderAlAwar, line 1107: I would rather use a for loop

pykokkos/lib/ufuncs.py
```diff
@@ -959,17 +960,166 @@ def multiply(viewA, viewB):
...
+
+@pk.workunit
+def subtract_impl_scalar_2d(team_member, cols, viewA, scalar, viewOut):
+    j: int = team_member.league_rank()
 
-def subtract(viewA, viewB):
+    def row_sub(i: int):
+        viewOut[j][i] = viewA[j][i] - scalar
+    
+    pk.parallel_for(pk.TeamThreadRange(team_member, cols), row_sub)
```
NaderAlAwar, line 1120: use a for loop

NaderAlAwar (changes requested):
Looks good overall, some changes required. One more thing I think we should do is move some of the broadcasting logic from `subtract()` to its own function since we will be using that in other ufuncs

pre_compile_tools/pre_compile_ufuncs.py
```diff
@@ -55,6 +55,9 @@ def main():
                     func_obj(v, v)
                 except (NotImplementedError, KeyError):
                     pass
+                except ValueError:
+                    if "broadcast" in func[0]:
+                        pass
```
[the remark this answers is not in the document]
HannanNaeem: Broadcast will throw a Value error, if the two views are not compatible. The array Api CI was failing because this exception was unhandled. I was not entirely sure what to do so, so I went with this. If you think we don't to skip this exception, I can add it to the skip. Let me know

pykokkos/lib/ufuncs.py
```diff
@@ -959,17 +960,166 @@ def multiply(viewA, viewB):
...
+
+@pk.workunit
+def stretch_fill_impl_2d(team_member, inner_its, col_wise, viewIn, viewOut):
+    tid: int = team_member.league_rank()
+    def inner_fill(i: int):
+        if col_wise:
+            viewOut[i][tid] = viewIn[i][0]
+        else:
+            viewOut[tid][i] = viewIn[0][i]
+    pk.parallel_for(pk.TeamThreadRange(team_member, inner_its), inner_fill)
```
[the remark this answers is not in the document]
HannanNaeem: Oh, I thought team policy added more parallelism by making the inner iterations run in parallel too, in contrast to a for loop where they would be be linear with every tid

pykokkos/lib/ufuncs.py
HannanNaeem: Let me give this a try, I thought view.shape returns a tuple and the constructor takes a list

pykokkos/lib/ufuncs.py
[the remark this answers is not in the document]
HannanNaeem: Changed to for loops

NaderAlAwar (approved):

PR #257, opened 2024-02-19
Fusion: allow fusion of a single parallel_reduce with a sequence of parallel_fors 

NaderAlAwar:
This PR enables fusion of a single parallel_reduce that with a sequence of parallel_fors that it follows.

### DIFF
--- a/pykokkos/core/fusion/fuse.py
+++ b/pykokkos/core/fusion/fuse.py
@@ -7,19 +7,27 @@ from .util import DeclarationsVisitor, VariableRenamer
 
 def fuse_workunit_kwargs_and_params(
     workunit_trees: List[ast.AST],
-    kwargs: Dict[str, Any]
+    kwargs: Dict[str, Any],
+    operation: str
 ) -> Tuple[Dict[str, Any], List[ast.arg]]:
     """
     Fuse the parameters and runtime arguments of a list of workunits and rename them as necessary
 
     :param workunits_trees: the list of workunit trees (ASTs) being merged
     :param kwargs: the keyword arguments passed to the call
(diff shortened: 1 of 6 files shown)


pykokkos/core/fusion/fuse.py
```diff
@@ -7,20 +7,28 @@
...
-    kwargs: Dict[str, Any]
+    kwargs: Dict[str, Any],
+    operation: str
 ) -> Tuple[Dict[str, Any], List[ast.arg]]:
     """
     Fuse the parameters and runtime arguments of a list of workunits and rename them as necessary
 
     :param workunits_trees: the list of workunit trees (ASTs) being merged
     :param kwargs: the keyword arguments passed to the call
+    :param operation: they type of parallel operation ("parallel_for", "parallel_reduce", or "parallel_scan")
```
gliga, line 18: they?

pykokkos/core/fusion/fuse.py
```diff
@@ -7,20 +7,28 @@
...
     :returns: a tuple of the fused kwargs and the combined inspected parameters
     """
 
+    if operation == "parallel_scan":
```
gliga, line 22: would be nice to put supported operations in an enum or just constants

pykokkos/interface/views.py
```diff
@@ -140,6 +140,9 @@ def __setitem__(self, key: Union[int, TeamMember], value: Union[int, float]) ->
         :param value: the new value at the index.
         """
 
+        if "PK_FUSION" in os.environ:
```
gliga, line 143: I love env variables

gliga (commented):
LG

PR #263, opened 2024-03-12
Add optimization that automatically adds the restrict keyword to kernels

NaderAlAwar:
This adds a feature (enabled by setting the `PK_RESTRICT` environment variable) to automatically add [`restrict`](https://en.cppreference.com/w/c/language/restrict) to views, which tells the compiler that the restricted views/arrays do not alias, enabling it optimize the code better. The alias analysis is done automatically by PyKokkos.

Ideally we would use the `Kokkos::Restrict` memory trait, however most compilers ignore `restrict` with pointers that are member variables (as it is with Kokkos Views). This PR solves this issue by generating a `KOKKOS_FUNCTION` that accepts the raw view pointers as arguments which we can then use `restrict` with.

### DIFF
--- a/pykokkos/core/fusion/util.py
+++ b/pykokkos/core/fusion/util.py
@@ -52,3 +52,28 @@ class VariableRenamer(ast.NodeTransformer):
         # If the name is not mapped, keep the original name
         node.arg = self.name_map.get(key, node.arg)
         return node
+
+def add_parent_refs(classdef: ast.ClassDef) -> ast.ClassDef:
+    """
+    Add references to each node's parent node in classdef (Copied from
+    static.py, can't be imported due to circular import)
+
+    :param classdef: the classdef being modified
+    :returns: the modified classdef
+    """
+
+    for node in ast.walk(classdef):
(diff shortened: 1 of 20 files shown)


pykokkos/core/fusion/util.py
```diff
@@ -52,3 +52,28 @@ def visit_keyword(self, node: ast.keyword) -> None:
         # If the name is not mapped, keep the original name
         node.arg = self.name_map.get(key, node.arg)
         return node
+
+def add_parent_refs(classdef: ast.ClassDef) -> ast.ClassDef:
+    """
+    Add references to each node's parent node in classdef (Copied from
+    static.py, can't be imported due to circular import)
```
gliga, line 59: could it be moved to a util file and then imported in both places?

gliga (approved):
LG (I like the tests ;)

pykokkos/core/fusion/util.py
NaderAlAwar: Yes, would require some more refactoring. I'll get to this in a later PR

PR #277, opened 2024-07-07
Add support for passing pytorch tensors directly to kernel calls

NaderAlAwar:
This checks if torch is installed and uses `torch.is_tensor()` to check if the passed argument is a pytorch tensor.

### DIFF
--- a/pykokkos/interface/parallel_dispatch.py
+++ b/pykokkos/interface/parallel_dispatch.py
@@ -121,6 +121,7 @@ def convert_arrays(kwargs: Dict[str, Any]) -> None:
     """
 
     cp_available: bool
+    torch_available: bool
 
     try:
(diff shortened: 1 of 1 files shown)


gliga:
If `torch` is not available, will it print a user friendly error message?

NaderAlAwar:
Not at the moment but this is not restricted to PyTorch. We need to have better error messages for invalid arguments passed to kernels.

IvanGrigorik:
Should be fine now @JBludau @gliga

JBludau (approved):
looks good other than the file ending

IvanGrigorik:
> looks good other than the file ending



> As discussed in another PR, can we please add CI check for all line endings

https://github.com/kokkos/pykokkos/pull/315

PR #280, opened 2024-07-15
Fix TeamPolicy constructor docstring.

kennykos:
TeamPolicy constructor docstring incorrectly copy and pasted from RangePolicy constructor. Update to reflect true description of league size and team size.


NaderAlAwar (approved):
Looks good, thanks!

PR #281, opened 2024-07-18
Add safety check for fusion

NaderAlAwar:
Added a safety check to prevent fusion of kernels that could introduce a dependency between threads in the fused kernel. Dependencies can only come from views, so the safety check looks at the what indices are used to access the views. Two kernels which access views only through their thread ID will not introduce a dependency between threads in the fused kernel.

### DIFF
--- a/pykokkos/core/fusion/trace.py
+++ b/pykokkos/core/fusion/trace.py
@@ -6,7 +6,7 @@ from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
 from pykokkos.core.parsers import Parser, PyKokkosEntity
 from pykokkos.interface import ExecutionPolicy, RangePolicy, ViewType
 
-from .access_modes import AccessMode, get_view_access_modes
+from .access_modes import AccessIndex, AccessMode, get_view_access_modes, get_view_write_indices_and_modes
 from .future import Future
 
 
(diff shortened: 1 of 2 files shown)


pykokkos/core/fusion/trace.py
```diff
@@ -189,6 +222,56 @@ def fuse(self, operations: List[TracerOperation], strategy: str) -> List[TracerO
 
         raise RuntimeError(f"Unrecognized fusion strategy '{strategy}'")
 
+    def is_safe_to_fuse(self, current: List[TracerOperation], current_views: Set[ViewType], current_safety_info: Dict[Tuple[int, int], Tuple[AccessIndex, AccessMode]], next: TracerOperation, next_views: Set[ViewType]) -> bool:
+        """
+        Check whether the next operation is safe to fuse with the
+        current operations
```
gliga, line 228: this comment should document the conditions under which it would be safe to fuse

pykokkos/core/fusion/trace.py
```diff
@@ -305,7 +458,7 @@ def fuse_operations(self, operations: List[TracerOperation]) -> TracerOperation:
             # Avoid long names
             fused_name = "_".join(names[:5]) + hashlib.md5(("".join(names)).encode()).hexdigest()
 
-        return TracerOperation(None, future, fused_name, policy, workunits, operation, parsers, fused_name, args, dependencies)
+        return TracerOperation(None, future, fused_name, policy, workunits, operation, parsers, fused_name, args, dependencies, fused_safety_info)
```
gliga, line 461: some tests for the newly introduced checks?

gliga (commented):
would be easier to review if some tests were there

PR #294, opened 2024-11-15
Add dependencies and pre-installation section

IvanGrigorik:
This PR adds pre-installation steps that could help users to install and use PyKokkos more smoothly.


IvanGrigorik:
@gliga

PR #297, opened 2024-12-18
Interface: add support for mathematical special functions (real inputs only)

### LINKED ISSUE #296: Interface: add cyl_bessel_i and cyl_bessel_i0 as mathemtical functions inside kernels (only accept real values for now)
Add support for the [regular modified cylindrical Bessel function](https://en.wikipedia.org/wiki/Bessel_function#Modified_Bessel_functions:_I.CE.B1_.2C_K.CE.B1) defined in cpp reference [https://en.cppreference.com/w/cpp/numeric/special_functions/cyl_bessel_i](https://en.cppreference.com/w/cpp/numeric/special_functions/cyl_bessel_i).

kennykos:
Add support for mathematical special functions defined in the header [https://github.com/kokkos/kokkos/blob/9fa2a01747f0a30ca2723c8d7d0a22c95a05717a/core/src/Kokkos_MathematicalSpecialFunctions.hpp#L4](https://github.com/kokkos/kokkos/blob/9fa2a01747f0a30ca2723c8d7d0a22c95a05717a/core/src/Kokkos_MathematicalSpecialFunctions.hpp#L4).

The bessel functions pk.cyl_bessel_XX() are only supported for real data types as complex numbers are not yet supported in pykokkos.


kennykos:
Duplicate of #359

PR #312, opened 2025-11-14
Update CI python version matrix to 3.10-3.13

kennykos:
Python versions now aligns with pykokkos-base.

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


kennykos:
@JBludau, @IvanGrigorik the CI is failing with python 3.13. This is in line with results on my local machine. 
The error being thrown is `RuntimeError: lib/ or lib64/ directories not found in installed pykokkos-base package. Try setting PK_KOKKOS_LIB_PATH instead.` This makes me think it's an `scikit-build` issue, but I'm not sure. The `cmake-install `directory is not being copied to `pykokkos-base.egg-info`, which is unexpected. However, when I copy over the cmake-install directory manually I get the same error.

gliga (approved):

PR #313, opened 2025-11-17
CI: add seperate workflows to build & import pykokkos on linux and macos

kennykos:
I was running into issues building pykokkos on my macos laptop, so I thought I'd open a PR. I think it's worthwhile to have these workflows separate, we may very well need differing install instructions for different OS's.

### DIFF
--- a/pykokkos/interface/execution_space.py
+++ b/pykokkos/interface/execution_space.py
@@ -13,6 +13,11 @@ class ExecutionSpace(Enum):
     Default = "Default"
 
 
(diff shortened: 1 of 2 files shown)


kennykos:
It looks like there's a possible solution on stack-overflow: [https://stackoverflow.com/a/79140822/22996414](https://stackoverflow.com/a/79140822/22996414), however, I'm not sure what cmake file to add this to.

IvanGrigorik:
@kennykos  do we still need this?
https://github.com/kokkos/pykokkos/pull/338 this one introduces ARM builds + testing

PR #320, opened 2025-12-15
algorithms: bounds

IvanGrigorik:
Add lower and upper bound using binary search
Tested for CUDA and OpenMP

API similar to standard `thrust::upper_bound`: https://nvidia.github.io/cccl/thrust/api/group__binary__search_1gac85cc9ea00f4bdd8f80ad25fff16741d.html
```python
# Thrust:
thrust::upper_bound(ForwardIterator first, ForwardIterator last, const LessThanComparable &value ) # -> return iterator
thrust::lower_bound(ForwardIterator first, ForwardIterator last, const GreaterThanComparable &value ) # -> return iterator

# Ours:
pk.upper_bound(view: pk.View, lastIndex: int, lessThanValue: int) # -> return int
pk.lower_bound(view: pk.View, lastIndex: int, greaterThanValue: int) # -> return int
```

Same for `lower_bound`
One of the student asked to implement this in order to use it from pykokkos workunits, since they needed this operation in their research


IvanGrigorik:
Updated description

examples/kokkos/lower_bound_example.py
```diff
@@ -0,0 +1,75 @@
...
+    # Find lower bound in scratch memory
+    bound_idx: int = pk.lower_bound(scratch, team_size, search_value)
+    
+    # Store result
+    result_view[globalIdx] = bound_idx
+
+
+# Test lower_bound with regular view
+@pk.workunit
+def lower_bound_view(i: int, view: pk.View1D[int], result_view: pk.View1D[int]):
```
gliga: no need for type annotations

examples/kokkos/upper_bound_example.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # Copy data to scratch and make it sorted within the team
+    scratch[team_rank] = view[globalIdx]
+    team_member.team_barrier()
+
+    # Now use upper_bound to find position in scratch
+    # For example, find upper bound for the value at team_rank position
+    search_value: int = team_rank * 2  # Search for a value
+    
+    # Find upper bound in scratch memory
```
gliga: unnecessary comment

examples/kokkos/upper_bound_example.py
```diff
@@ -0,0 +1,75 @@
...
+    bound_idx: int = pk.upper_bound(scratch, team_size, search_value)
+    
+    # Store result
+    result_view[globalIdx] = bound_idx
+
+
+# Test upper_bound with regular view
+@pk.workunit
+def upper_bound_view(i: int, view: pk.View1D[int], result_view: pk.View1D[int]):
```
gliga: as before

gliga (commented):
Is it sufficient to work only with ints?

examples/kokkos/upper_bound_example.py
```diff
@@ -0,0 +1,75 @@
...
+
+    print(f"Total elements: {N}, Team size: {team_size}, Number of teams: {num_teams}")
+    print(f"Initial view: {view}")
+
+    # Test with TeamPolicy (scratch memory)
+    team_policy = pk.TeamPolicy(pk.ExecutionSpace.Cuda, num_teams, team_size)
+    
+    print("\nRunning upper_bound with scratch memory...")
+    pk.parallel_for(team_policy, team_upper_bound, view=view, result_view=result_view)
+    print(f"Result (scratch upper_bound): {result_view}")
```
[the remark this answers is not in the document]
IvanGrigorik: Added

examples/kokkos/upper_bound_example.py
IvanGrigorik: Removed

IvanGrigorik:
For some reason tests fails due unknown reason. I couldn't reproduce but on my local machine nor on our local server
@kennykos was able to reproduce issue, but I cant help him, since it's on the server that I can not access

IvanGrigorik:
Closing. 
It should not be translated like that. This PR builds an AST and translates it to C++. 
Instead, we need to create a general structure and pass the `algorithm` class to it, which will define its behavior.

PR #329, opened 2025-12-19
core: scratch size getting for multiple types and dimensions

IvanGrigorik:
Added unit tests and implementation of `ScratchView?D` (where `?` is dimension from 1 to 8  and `shmem_size` to get shared memory size for any dimension and for any type


gliga (approved):
LG

Issue #330, opened 2025-12-22
Array API CI is failing


IvanGrigorik:
[![Array API Testing](https://github.com/kokkos/pykokkos/actions/workflows/array_api.yml/badge.svg)](https://github.com/kokkos/pykokkos/actions/workflows/array_api.yml)

I saw that we have a problem with `array_api` workflow in the Docker pull request https://github.com/kokkos/pykokkos/pull/305

Since I haven't modified any PyKokkos code itself it was a surprise to find it out. I decided to re-run main branch CI for array_api testing, which was not failing before rerun https://github.com/kokkos/pykokkos/actions/runs/20354112743. Now, it is failing.

I assume this is related to some dependency or array_api itself.

PR #332, opened 2025-12-22
views: change 0D arrays conversion (array_api fail)

IvanGrigorik:
This PR fixes `views` incompatibility with 0D arrays, which cause new array_api check (as I guess) fail in main workflow.
Closes https://github.com/kokkos/pykokkos/issues/330


gliga:
> Added unit tests based on @kennykos script.

That is great

PR #337, opened 2025-12-26
build: fix CUDA_HOME env variable

IvanGrigorik:
This PR fixes `CUDA_HOME` environment variable
Now, users need to to specify it explicitly to build it
This fixes that

### DIFF
--- a/base/cmake/Modules/KokkosPythonKokkos.cmake
+++ b/base/cmake/Modules/KokkosPythonKokkos.cmake
@@ -83,6 +83,7 @@ IF(_INTERNAL_KOKKOS)
     SET(OpenMP_FOUND OFF)
     SET(Threads_FOUND OFF)
     SET(CUDA_FOUND OFF)
+    SET(HIP_FOUND OFF)
 
     # Enable OpenMP if explicitly requested
     IF((DEFINED ENABLE_OPENMP AND ENABLE_OPENMP) OR (DEFINED Kokkos_ENABLE_OPENMP AND Kokkos_ENABLE_OPENMP))
@@ -119,6 +120,17 @@ IF(_INTERNAL_KOKKOS)
         SET(CUDA_FOUND ON)
     ENDIF()
 
+    # search for HIP if explicitly enabled
+    IF((DEFINED ENABLE_HIP AND ENABLE_HIP) OR (DEFINED Kokkos_ENABLE_HIP AND Kokkos_ENABLE_HIP))
+        INCLUDE(CheckLanguage)
+        CHECK_LANGUAGE(HIP)
+        
+        IF(CMAKE_HIP_COMPILER)
+            ENABLE_LANGUAGE(HIP)
+            SET(HIP_FOUND ON)
+        ENDIF()
+    ENDIF()
+
     ADD_OPTION(ENABLE_SERIAL "Enable Serial backend when building Kokkos submodule" ON)
     # If OpenMP was explicitly requested, use that; otherwise use auto-detection result
     IF((DEFINED ENABLE_OPENMP AND ENABLE_OPENMP) OR (DEFINED Kokkos_ENABLE_OPENMP AND Kokkos_ENABLE_OPENMP))
@@ -134,6 +146,8 @@ IF(_INTERNAL_KOKKOS)
     ENDIF()
     # CUDA must be explicitly enabled - default to OFF
     ADD_OPTION(ENABLE_CUDA "Enable CUDA when building Kokkos submodule" OFF)
+    # HIP must be explicitly enabled - default to OFF
+    ADD_OPTION(ENABLE_HIP "Enable HIP when building Kokkos submodule" OFF)
 
     # always disable pthread backend since pthreads are not supported on Windows
     IF(WIN32)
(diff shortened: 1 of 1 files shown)


kennykos:
> This PR fixes `CUDA_HOME` environment variable Now, users need to to specify it explicitly to build it This fixes that

How was the `CUDA_HOME` environment variable broken? What is the specific bug that this is addressing?

base/cmake/Modules/KokkosPythonKokkos.cmake
```diff
@@ -134,6 +292,21 @@ IF(_INTERNAL_KOKKOS)
...
     ENDIF()
 
+    # Enforce mutual exclusion AFTER syncing from command-line options
+    # Kokkos doesn't allow both OpenMP and Threads to be enabled
+    IF(ENABLE_OPENMP AND ENABLE_THREADS)
+        # Prefer the explicitly set option, or OpenMP if both are explicit
+        IF(DEFINED Kokkos_ENABLE_THREADS AND NOT DEFINED Kokkos_ENABLE_OPENMP)
+            SET(ENABLE_OPENMP OFF)
+            SET(Kokkos_ENABLE_OPENMP OFF CACHE BOOL "Build Kokkos submodule with OpenMP support" FORCE)
+            MESSAGE(STATUS "Disabling OpenMP because Threads was explicitly enabled")
```
gliga: would it be better to fail?

base/cmake/Modules/KokkosPythonKokkos.cmake
```diff
@@ -84,26 +84,184 @@ IF(_INTERNAL_KOKKOS)
...
+
+            # Add CUDA include directories to the include path
+            INCLUDE_DIRECTORIES(SYSTEM ${CUDAToolkit_INCLUDE_DIRS})
+        ELSE()
+            SET(CUDA_FOUND OFF)
+        ENDIF()
+    ELSEIF(_CUDA_EXPLICITLY_ENABLED)
+        # CUDA explicitly enabled - ensure we have it
+        # Search for CUDA in common installation directories
+        FILE(GLOB CUDA_SEARCH_PATHS
```
gliga: is this code duplicate what we already had above?

base/cmake/Modules/KokkosPythonKokkos.cmake
IvanGrigorik: Add a specific check if OpenMP and Threads enabled

base/cmake/Modules/KokkosPythonKokkos.cmake
```diff
@@ -84,26 +84,114 @@ IF(_INTERNAL_KOKKOS)
...
+    ENDIF()
+
+    # Only search for CUDA if explicitly enabled
+    IF(_CUDA_EXPLICITLY_ENABLED)
+        FILE(GLOB CUDA_SEARCH_PATHS
+            "/usr/local/cuda-*"
+            "/usr/local/cuda"
+            "/opt/cuda-*"
+            "/opt/cuda"
+        )
```
JBludau: This is bad practice. We should just do a find package.

base/cmake/Modules/KokkosPythonKokkos.cmake
```diff
@@ -84,26 +84,114 @@ IF(_INTERNAL_KOKKOS)
...
+                    ENDIF()
+                ENDFOREACH()
+            ENDIF()
+
+            # Set up env for this build
+            SET(ENV{CUDA_HOME} "${CUDA_TOOLKIT_ROOT}")
+            SET(ENV{CUDACXX} "${CUDAToolkit_NVCC_EXECUTABLE}")
+            IF(NOT CMAKE_CUDA_COMPILER)
+                SET(CMAKE_CUDA_COMPILER "${CUDAToolkit_NVCC_EXECUTABLE}" CACHE FILEPATH "CUDA compiler" FORCE)
+            ENDIF()
```
JBludau: This is also against CMakes practice ...

base/cmake/Modules/KokkosPythonKokkos.cmake
```diff
@@ -84,26 +84,114 @@ IF(_INTERNAL_KOKKOS)
...
+
+            # Enable CUDA language now that we have the correct compiler
+            INCLUDE(CheckLanguage)
+            CHECK_LANGUAGE(CUDA)
+            IF(CMAKE_CUDA_COMPILER)
+                ENABLE_LANGUAGE(CUDA)
+            ENDIF()
```
JBludau: In CMake packages should not enable languages, but work with the languages given by the user or error out. Nevertheless, for something like Kokkos, this does not really work...so this would be ok here ... but we should make it consistent for HIP, too

base/cmake/Modules/KokkosPythonKokkos.cmake
```diff
@@ -134,6 +222,21 @@ IF(_INTERNAL_KOKKOS)
...
+            SET(ENABLE_OPENMP OFF)
+            SET(Kokkos_ENABLE_OPENMP OFF CACHE BOOL "" FORCE)
+            MESSAGE(STATUS "Disabling auto-detected OpenMP (Threads explicitly enabled)")
+        ELSE()
+            SET(ENABLE_THREADS OFF)
+            SET(Kokkos_ENABLE_THREADS OFF CACHE BOOL "" FORCE)
+            MESSAGE(STATUS "Disabling Threads (OpenMP enabled)")
+        ENDIF()
+    ENDIF()
+
```
JBludau: Why are we repeating logic that is handled in Kokkos Core? Users should get a pretty descriptive message about this

JBludau (changes requested):
I don't think that is going in the right direction

base/cmake/Modules/KokkosPythonKokkos.cmake
```diff
@@ -84,26 +84,114 @@ IF(_INTERNAL_KOKKOS)
     SET(Threads_FOUND OFF)
     SET(CUDA_FOUND OFF)
 
-    IF(NOT Kokkos_ENABLE_THREADS)
+    # Only search for OpenMP if not explicitly disabled and Threads not explicitly enabled
+    IF(NOT DEFINED Kokkos_ENABLE_THREADS OR NOT Kokkos_ENABLE_THREADS)
```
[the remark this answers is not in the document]
IvanGrigorik: Sounds good. This way we dont need to do later checks

base/cmake/Modules/KokkosPythonKokkos.cmake
IvanGrigorik: I got error before openmp fix. Now everything is fine

base/cmake/Modules/KokkosPythonKokkos.cmake
IvanGrigorik: Added HIP support.

JBludau:
> @JBludau I have some issues with pykokkos build now, and it looks like its related to Kokkos itself:
> 
> ```
> [  1%] Building CXX object _deps/kokkos-build/simd/src/CMakeFiles/kokkossimd.dir/Kokkos_SIMD_dummy.cpp.o
> [  2%] Linking CXX shared library libkokkossimd.so
> [  2%] Built target kokkossimd
> [  3%] Building CXX object _deps/kokkos-build/core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_Abort.cpp.o
> [  4%] Building CXX object _deps/kokkos-build/core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_CPUDiscovery.cpp.o
> [  5%] Building CXX object _deps/kokkos-build/core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_Command_Line_Parsing.cpp.o
> [  6%] Building CXX object _deps/kokkos-build/core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_Core.cpp.o
> ~/pykokkos/base/_skbuild/linux-x86_64-3.12/cmake-build/_deps/kokkos-src/core/src/Kokkos_MathematicalFunctions.hpp(301): warning #20013-D: calling a constexpr __host__ function("remainder") from a __host__ __device__ function("remainder") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.
> 
> Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"
> 
> ~/pykokkos/base/_skbuild/linux-x86_64-3.12/cmake-build/_deps/kokkos-src/core/src/Kokkos_MathematicalFunctions.hpp(301): warning #20013-D: calling a constexpr __host__ function("remainder") from a __host__ __device__ function("remainderf") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.
> ```
> 
> And these checks repeated multiple times. AFAIK, this is the issue with CUDA declarations, but I don't understand why this issue arouse only after my changes.

That is weird ... let me see if I can track that down

PR #340, opened 2026-01-02
Recompile pk_cpp if outdated

### LINKED ISSUE #234: pk_cpp is not removed if kernel changes which keeps reporting an error
The following example leads to an error:

```
import pykokkos as pk

@pk.workunit
def map_d(i):
    #i += 1                                                                                                                                                                                                                                                                                                                                           
    pass

def main():
    SIZE = 2
    pk.parallel_for(SIZE, map_d)

main()
```

This happens as we do not support `pass` (but that is a side comment only). The key is that this results in an error.

Once I "fix" the kernel and have the resulting (correct) code:
```
import pykokkos as pk

@pk.workunit
def map_d(i):
    i += 1

def main():
    SIZE = 2
    pk.parallel_for(SIZE, map_d)

main()
```

I am still getting the same error as before.  Only after deleting `pk_cpp` in between kernel updates, the error goes away and my example runs properly.

Action: Remove pk_cpp when kernel changes.

Used pykokkos: `2e0b664c2aec1c923c55a7dc5958fc19f0c327cd`
Used command: `./pk pk_example X.py`

kennykos:
Closes #234

### DIFF
--- a/pykokkos/core/compiler.py
+++ b/pykokkos/core/compiler.py
@@ -173,35 +173,6 @@ class Compiler:
         }:
             raise Exception(f"Types are required for style: {entity.style}")
 
-        if self.is_compiled(module_setup.output_dir):
-            if hash not in self.members:  # True if pre-compiled
-                if len(metadata) > 1:
-                    entity, classtypes = self.fuse_objects(
-                        metadata, fuse_ASTs=True, **kwargs
-                    )
-
-                if types_inferred:
-                    entity.AST = parser.fix_types(entity, updated_types)
-                if decorator_inferred:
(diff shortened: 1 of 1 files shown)


IvanGrigorik:
@kennykos can you add some unit test to verify the correctness?

IvanGrigorik:
> > @kennykos can you add some unit test to verify the correctness?
> 
> I'm not sure how to write unit tests for this without using `exec` as it involves files being changed. Do you have any ideas?

It would be enough to test `is_current` directly, not whole app.
But if you want to check the whole app - you can append one char of file during tests and do 3 exec rounds in a loop or something like that

Issue #342, opened 2026-01-05
[error] pykokkos function breaks with undiscernible traceback


kennykos:
I want a pykokkos helper function that calculates the lagrange basis for a value at four grid points. 

However, my code

```
import numpy as np
import pykokkos as pk

@pk.function
def lagrange_basis(x, xi, xj, xk, xl) -> pk.double:
    """Calculate the Lagrange basis polynomial."""
    value = (x - xj) * (x - xk) * (x - xl) / ((xi - xj) * (xi - xk) * (xi - xl))
    return value

@pk.workunit
def work(wid):
    lagrange_basis(1,0,0,0,0)

def main():
    N = 10
    pk.parallel_for("work", 10, work)

main()
```

breaks with the traceback

```
Traceback (most recent call last):
  File "/work/09661/gkk345/vista/tmp/pykokkos_lagrange.py", line 18, in <module>
    main()
    ~~~~^^
  File "/work/09661/gkk345/vista/tmp/pykokkos_lagrange.py", line 16, in main
    pk.parallel_for("work", 10, work)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/work/09661/gkk345/vista/pykokkos/pykokkos/interface/parallel_dispatch.py", line 184, in parallel_for
    runtime_singleton.runtime.run_workunit(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        handled_args.name, handled_args.policy, handled_args.workunit, "for", **kwargs
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/work/09661/gkk345/vista/pykokkos/pykokkos/core/runtime.py", line 193, in run_workunit
    return self.execute_workunit(
           ~~~~~~~~~~~~~~~~~~~~~^
        name, policy, workunit, operation, parser, **kwargs
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/work/09661/gkk345/vista/pykokkos/pykokkos/core/runtime.py", line 253, in execute_workunit
    members: PyKokkosMembers = self.precompile_workunit(
                               ~~~~~~~~~~~~~~~~~~~~~~~~^
        workunit,
[...]

PR #344, opened 2026-01-05
functions: type autoinference

IvanGrigorik:
This PR addresses https://github.com/kokkos/pykokkos/issues/342.

Fixed type inference for functions, so now no need to specify argument types.

### DIFF
--- a/pykokkos/core/translators/static.py
+++ b/pykokkos/core/translators/static.py
@@ -353,7 +353,9 @@ class StaticTranslator:
             if not types_changed:
                 break
         else:
-            print(f"Warning: Type inference did not converge after {max_iterations} iterations")
+            print(
+                f"Warning: Type inference did not converge after {max_iterations} iterations"
+            )
 
     def translate_functions(
         self, source: Tuple[List[str], int], restrict_views: Set[str]


IvanGrigorik:
@kennykos can you please check and confirm that it fixes https://github.com/kokkos/pykokkos/issues/342?

You have another problem with [this](https://github.com/kokkos/pykokkos/issues/342#:~:text=value%20%3D%20%28x%20%2D%20xj%29%20%2A%20%28x%20%2D%20xk%29%20%2A%20%28x%20%2D%20xl%29%20%2F%20%28%28xi%20%2D%20xj%29%20%2A%20%28xi%20%2D%20xk%29%20%2A%20%28xi%20%2D%20xl%29%29) line. 
Pykokkos requires specifying the types of new values, since autotype inference for all values can lead to some problems (we can fix it in the future, but this PR addresses only the argument inference part.

pykokkos/core/visitors/kokkosfunction_visitor.py
kennykos: This will still break if the return type is not specified. We want to do something like `node.returns`, but if this is `None` then `print_pretty` will break in `visitor_utils.py`.

kennykos:
> @kennykos can you please check and confirm that it fixes #342?
> 
> You have another problem with [this](https://github.com/kokkos/pykokkos/issues/342#:~:text=value%20%3D%20%28x%20%2D%20xj%29%20%2A%20%28x%20%2D%20xk%29%20%2A%20%28x%20%2D%20xl%29%20%2F%20%28%28xi%20%2D%20xj%29%20%2A%20%28xi%20%2D%20xk%29%20%2A%20%28xi%20%2D%20xl%29%29) line. Pykokkos requires specifying the types of new values, since autotype inference for all values can lead to some problems (we can fix it in the future, but this PR addresses only the argument inference part.

This fixes #342, given my caveat above about the mangled error message when the return type is not specified.

pykokkos/core/visitors/kokkosfunction_visitor.py
IvanGrigorik: Can you please merge your branch into mine? since I remember you said that you fixed that

pykokkos/core/translators/static.py
```diff
@@ -230,6 +237,116 @@ def translate_functions(
...
+                        # if parameter name is a common thread ID name, default to int
+                        if not found_type and param_name in (
+                            "tid",
+                            "i",
+                            "j",
+                            "k",
+                            "idx",
+                            "index",
+                        ):
+                            inferred_types[param_name] = "int"
```
kennykos: What is the justification for this? Why won't the logic above be able to infer the type? 
I'm not a huge fan on this. For instance, some examples use "wid" which is not covered here. Can we delete?

pykokkos/core/type_inference/args_type_inference.py
```diff
@@ -56,10 +56,53 @@ class UpdatedDecorator:
...
+    :param prefer_float: If True and value is int, prefer double over int
+    :returns: Type string in the format used by type inference (e.g., "int", "double", "numpy:int64")
+    """
+    param_type = type(value).__name__
+
+    if param_type == "int":
+        if value.bit_length() > 31:
+            param_type = f"numpy:{DataType.int64.name}"
+        elif prefer_float:
+            param_type = DataType.double.name
```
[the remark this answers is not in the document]
IvanGrigorik: I used this approach since PyKokkos failed on some functions like that:
```python
@pk.function
def some_calc(x) -> float:
    return x * 2.5

# 5 as int, but function returns float
result = self.some_calc(5)
```

Now everything should be alright

pykokkos/core/type_inference/args_type_inference.py
```diff
@@ -56,10 +56,53 @@ class UpdatedDecorator:
     "float32",
 ]
 
-# Cache for original argument nodes: Maps stringified workunit reference, e.g str(workunit_name), to the original ast.arguments node
+# Supported array libraries for type inference
+SUPPORTED_ARRAY_LIBRARIES = ("numpy", "cupy", "torch", "jax", "jaxlib")
```
kennykos: great!

kennykos (approved):
Looks good to me! I left one comment about an error message that could be nice, but I'll defer to @IvanGrigorik on that one. Otherwise, looks good to me!

PR #345, opened 2026-01-12
tests: workload deprecation

IvanGrigorik:
This PR deprecates workload functions in tests. 
Next PRs will deprecate this function in pykokkos itself

The only test that we deprecated entirely is `test_lambdas`, since lambdas can not be used without workloads.
I can work on it, but it should be a dedicated task itself, since it looks not straightforward

### DIFF
--- a/tests/test_lambdas.py
+++ b/tests/test_lambdas.py
@@ -1,3 +1,5 @@
+#  # for now this test is deprecated and should be translated after closing this issue:
+#  # https://github.com/kokkos/pykokkos/issues/346
 # import unittest
 
 # import pykokkos as pk


tests/test_AST_translator.py
```diff
@@ -184,10 +184,6 @@ def test_math_constants(self):
 
         assert_allclose(expected_result, result)
 
-    # def test_constants(self):
```
gliga: do we have equivalent tests for standalone mode?

IvanGrigorik: We can have some equivalent for this.
We can return constants value accumulator from function and compare it to local value.

IvanGrigorik: Added

tests/test_parallelfor.py
```diff
@@ -1,44 +0,0 @@
-# import unittest
-# import pykokkos
```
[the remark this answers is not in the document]
IvanGrigorik: This testcase was deprecated before my commit, so its just a cleanup.

PR #349, opened 2026-01-20
build: add release link instead of branch

IvanGrigorik:
Previously, we used the release candidate branch.
It got removed
I fixed it by using the release itself, not the candidate

### DIFF
--- a/.github/workflows/formatting.yml
+++ b/.github/workflows/formatting.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install black
+          pip install black==25.12.0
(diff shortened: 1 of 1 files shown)


.github/workflows/formatting.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install black
+          pip install black==25.12.0
```
JBludau: That seems unrelated?

JBludau: Is there a reason for why this is here?

JBludau (approved):
The rest is fine

Issue #348, opened 2026-01-16
Translating List types


kennykos:
Pykokkos is currently unable to translate the following script:

```py
import numpy as np
import pykokkos as pk
from typing import List

@pk.workunit
def work(wid, a, l:List[int]):
    a[wid] = a[wid] + l[wid]

def main():
    N = 10
    a = np.ones(N)
    l = [1] * 10
    print(l)
    # ... do anything with a using numpy
    pk.parallel_for("work", 10, work, a=a, l=l)
    print(a)

main()
```

It fails with the traceback

```py
Traceback (most recent call last):
  File "/work/09661/gkk345/vista/tmp/pk_list.py", line 18, in <module>
    main()
    ~~~~^^
  File "/work/09661/gkk345/vista/tmp/pk_list.py", line 15, in main
    pk.parallel_for("work", 10, work, a=a, l=l)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/09661/gkk345/vista/pykokkos/pykokkos/interface/parallel_dispatch.py", line 184, in parallel_for
    runtime_singleton.runtime.run_workunit(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        handled_args.name, handled_args.policy, handled_args.workunit, "for", **kwargs
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/work/09661/gkk345/vista/pykokkos/pykokkos/core/runtime.py", line 193, in run_workunit
    return self.execute_workunit(
           ~~~~~~~~~~~~~~~~~~~~~^
        name, policy, workunit, operation, parser, **kwargs
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/work/09661/gkk345/vista/pykokkos/pykokkos/core/runtime.py", line 253, in execute_workunit
    members: PyKokkosMembers = self.precompile_workunit(
                               ~~~~~~~~~~~~~~~~~~~~~~~~^
        workunit,
        ^^^^^^^^^
    ...<6 lines>...
        **kwargs,
        ^^^^^^^^^
    )
    ^
[...]

PR #354, opened 2026-01-21
core: add dynamic type inference based on workunit types

IvanGrigorik:
This PR fixes the issue https://github.com/kokkos/pykokkos/issues/348
The problem was that when you wrote `l: List[int]`, Python created an `ast.Name` object, and accessing `.attr` on it caused the crash.

Now we handle this edge case as well.

### DIFF
--- a/pykokkos/core/visitors/parameter_visitor.py
+++ b/pykokkos/core/visitors/parameter_visitor.py
@@ -57,6 +57,26 @@ class ParameterVisitor(ast.NodeVisitor):
         for a in args:
             self.visit(a)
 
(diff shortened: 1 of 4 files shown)


pykokkos/core/visitors/parameter_visitor.py
```diff
@@ -74,6 +74,13 @@ def visit_arg(self, node: ast.arg) -> None:
         if decltype is None:
             self.error(node, "Type is not supported")
 
+        if isinstance(annotation, ast.Subscript):
```
gliga: any comment here?

gliga (approved):
Mostly LG

PR #363, opened 2026-02-01
examples: update scratch-related examples and add cuda example

IvanGrigorik:
Updated our deprecated examples.
After merging https://github.com/kokkos/pykokkos/pull/321, we can specify scratch size for workunits using a standard function, and we can do that for CUDA.

I received a question on how to use scratch memory for CUDA, so the updated example should help
We also may consider adding a doc page about scratch (how it works, how to specify its size, etc.).

### DIFF
--- a/examples/kokkos/inclusive_scan_teamCuda.py
+++ b/examples/kokkos/inclusive_scan_teamCuda.py
@@ -1,3 +1,4 @@
+import cupy as cp
 import pykokkos as pk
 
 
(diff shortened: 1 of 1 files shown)


Wilson1211:
Using solely `s_g : pk.ScratchView1D[int] = pk.ScratchView1D(team_member.team_scratch(0), 3 * team_size)` in kernel function works in old version of pykokkos, but not anymore in new version.
Need to first declare variable scratch_size
`scratch_size: int = pk.ScratchView1D[int].shmem_size(3*team_size)`
then 
`pk.TeamPolicy 
 pk.parallel_for(pk.TeamPolicy(int(league_size), int(team_size)).set_scratch_size(0, pk.PerTeam(scratch_size)), load_balance`
 in host. After that, I am able to use `s_g : pk.ScratchView1D[int] = pk.ScratchView1D(team_member.team_scratch(0), 3 * team_size)` in kernel

Issue #361, opened 2026-01-30
ufuncs removal


IvanGrigorik:
This is a sub-issue of https://github.com/kokkos/pykokkos/issues/356.

We should remove all `ufunc` interface and functions that should **not** be called in the kernels. 
For example:

This ufunc can be called from kernel itself, so it should stay
```python
@pk.workunit
def compute_workunit(tid: int, view: pk.View1D[pk.double]) -> None:
    view[tid] = sqrt(view[tid])      # scalar sqrt

```

On the other side,

```python
view = pk.View([10], pk.double)
view[:] = np.arange(10, dtype=np.float64)

result1 = pk.sqrt(view)       # Square root of all elements
result2 = pk.log(view)        # Natural log of all elements  
result3 = pk.sin(view)        # Sine of all elements
result4 = pk.exp(view)        # Exponential of all elements
```

This ufunc **can not** be called from the kernel, so it should go. 
The main difference is with the working elements (scalars vs DS)

Here is the approximate list of ufuncs that should be inspected removed:

<details><summary>List of ufuncs</summary>
<p>

pk.reciprocal
pk.log
pk.log2
pk.log10
pk.log1p
pk.sqrt
pk.sign
pk.add
pk.copyto
pk.subtract
pk.dot
pk.multiply
pk.matmul
pk.np_matmul
pk.divide
pk.negative
pk.positive
pk.power
pk.fmod
pk.square
pk.greater
pk.logaddexp
pk.true_divide
pk.logaddexp2
pk.floor_divide
pk.sin
pk.cos
pk.tan
pk.tanh
pk.logical_and
pk.logical_or
pk.logical_xor
pk.logical_not
pk.fmax
pk.fmin
pk.exp
pk.exp2
pk.argmax
pk.unique
pk.var
pk.in1d
pk.mean
pk.hstack
pk.transpose
pk.index
pk.isinf
pk.isnan
pk.equal
pk.isfinite
pk.round
pk.trunc
pk.ceil
pk.floor
pk.broadcast_view
pk.reshape
pk.ravel
pk.expand_dims
pk.all
pk.any
pk.sum
pk.find_max
pk.searchsorted
pk.col
pk.linspace
pk.logspace
pk.zeros
pk.zeros_like
pk.ones
pk.ones_like
pk.full
pk.full_like
pk.iinfo
pk.finfo

</p>
</details>

PR #367, opened 2026-02-12
ufunc removal

nayyirahsan:
Deleted ufunc implementations and tests. Replaced ufunc usage in examples with NumPy. Scalar math in kernels remains. Adjusted __init__.py and views.py accordingly. Array API CI workflow kept but precompile ufuncs test is commented out.

Addresses issue #361

### DIFF
--- a/pykokkos/__init__.py
+++ b/pykokkos/__init__.py
@@ -17,7 +17,12 @@ from pykokkos.kokkos_manager import (
     set_device_id,
 )
 
-from pykokkos.lib.ufuncs import _isnan as isnan, _isinf as isinf, _isfinite as isfinite, _equal as equal
+from pykokkos.lib.ufuncs import (
+    _isnan as isnan,
+    _isinf as isinf,
+    _isfinite as isfinite,
+    _equal as equal,
+)
 
 from pykokkos.lib.info import iinfo, finfo
 from pykokkos.lib.create import zeros, zeros_like, ones, ones_like, full, full_like


IvanGrigorik:
Overall LG
You also need to change some array api tests here
https://github.com/kokkos/pykokkos/pull/367/changes#diff-de4346c5bea6a8c4efd61f7378aac35876f468d8486137fa77b336175f429f8dR60

For example, remove the `sqrt` function (the function that was removed by this PR)

pykokkos/lib/pk_array.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+
+
+class PKArray(np.ndarray):
```
IvanGrigorik: Our final goal is to remove any abstraction of pk.View-s  
Introducing a new abstraction is not a good idea in this case
Why do we need it, when we can just change the test expected value?

nayyirahsan: You're right, I reverted the PKArray implementation. The failing tests are for functions that delegate to numpy, and since numpy guarantees their correctness, I just removed those tests from the workflow instead.

pykokkos/interface/views.py
```diff
@@ -551,7 +550,9 @@ def __hash__(self):
     def __index__(self) -> int:
         return int(self.data[0])
 
-    def __array__(self, dtype=None):
+    def __array__(self, dtype=None, copy=None):
+        if self.shape == ():
+            return np.squeeze(self.data)
```
IvanGrigorik: This thing will mess up our code a lot.
We support NumPy, CuPy, Tensors, Python arrays, etc.
You got it.
I missed that in the previous review.
***We should not*** use numpy functions to perform any operations except numpy-oriented DS.
In this case, this View is an abstraction of all arrays mentioned above. If we want to support them all, we need to generalize all functions

---

My suggestion for such cases is not removing some functions (like equal), but moving them and making them private, so we would be able to use it inside our code base

IvanGrigorik (changes requested):
Ok there are some serious problems that I missed on my first review. 
If you have any access to Nvidia GPUs - please, try to test your changes on them using Cuda execution space (or run some existing [examples/tests](https://github.com/kokkos/pykokkos/blob/main/examples/pykokkos/from_array.py))

Unfortunately, we don't have any NVIDIA GitHub runners, and we can't verify the correctness of changes across different systems, so this burden lies on developers. :(

PR #381, opened 2026-02-27
core: multi-scratch specification

IvanGrigorik:
This PR addresses the issue of specifying multiple scratch sizes using updates decorator-based syntax.
Before, you could do:

```python
shmem_float1: int = pk.ScratchView1D[pk.float].shmem_size(<size_float1>)
shmem_int: int = pk.ScratchView1D[int].shmem_size(<size_int>)
shmem_float2: int = pk.ScratchView1D[pk.float].shmem_size(<size_float2>)
```
Since we deprecated this in order to make a scratch workunit-specific, we need to be able to set multiple scratch sizes in our `@pk.workunit(scratch=[...])` decorator as well.

With the new syntax, you can do it like this:
```python
@pk.workunit(
    scratch=[
        (int,   lambda p: p.n_ints),
        (float, lambda p: p.N),
    ]
)
def workunit(team_member: pk.TeamMember, N, n_ints, ...)
```

### DIFF
--- a/pykokkos/core/runtime.py
+++ b/pykokkos/core/runtime.py
@@ -119,6 +119,17 @@ def apply_scratch_spec(workunit: Callable, policy: TeamPolicy, **kwargs) -> None
         if total_scratch_size > 0:
             policy.scratch_size_level = 0
             policy.scratch_size_value = PerTeam(total_scratch_size)
+            try:
+                max_scratch = TeamPolicy.scratch_size_max(0)
+                if max_scratch > 0 and total_scratch_size > max_scratch:
+                    raise ValueError(
+                        f"Requested scratch size ({total_scratch_size} bytes) "
+                        f"exceeds maximum for level 0 ({max_scratch} bytes). "
+                        "Reduce scratch allocation or use a different level."
+                    )
+            except (ImportError, AttributeError):
+                # backend may not expose scratch_size_max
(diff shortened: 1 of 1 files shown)


pykokkos/core/runtime.py
kennykos: Is there any way to check if total scratch size will be out of bounds?

IvanGrigorik:
@kennykos can you approve this one?

pykokkos/core/runtime.py
IvanGrigorik: @kennykos  done.

pykokkos/core/runtime.py
```diff
@@ -93,16 +102,34 @@ def apply_scratch_spec(workunit: Callable, policy: TeamPolicy, **kwargs) -> None
...
+            else:
+                num_elements = size_func(policy)
             total_scratch_size += _calculate_aligned_scratch_size(dtype, num_elements)
 
         if total_scratch_size > 0:
             policy.scratch_size_level = 0
             policy.scratch_size_value = PerTeam(total_scratch_size)
+            try:
+                max_scratch = TeamPolicy.scratch_size_max(0)
+                if max_scratch > 0 and total_scratch_size > max_scratch:
```
kennykos: should we also throw an error if max_scratch == 0?

Issue #380, opened 2026-02-27
Translate C++ functor.hpp with clang-format (if available)


kennykos:
When developing pykokkos kernels it's not uncommon to get errors of the form

```
/work/09661/gkk345/vista/3dcapsules/python/sph_capsules/pk_cpp/sph/_pk_kernels/_local/_local_local_reg_stokes3d_near_correction_fp64/types_2bb60e03853aafec486471292efa709e/Cuda/../functor.hpp(9): error: operand types are incompatible ("double" and "double *")
[...]

PR #382, opened 2026-02-27
add clang-format option from env variable

IvanGrigorik:
As it was said  in this issue:
https://github.com/kokkos/pykokkos/issues/380#issuecomment-3975606152

We should have total flexibility over our tool, including formatting of intermediate files, if needed
This PR adds `PK_FORMAT` env variable dependency, so when `export PK_FORMAT=1` we will do `clang-format` pass on our IR files

### DIFF
--- a/docs/environment_variables.rst
+++ b/docs/environment_variables.rst
@@ -1,0 +1,48 @@
+
+Environment Variables
+=====================
+
+PyKokkos behavior can be controlled through the following environment variables:
+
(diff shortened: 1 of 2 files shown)


kennykos (changes requested):
Can we add documentation for the environment variable before this is merged in?

kennykos (approved):
Thanks! Looks good to me.

PR #385, opened 2026-02-28
core: binding fix

IvanGrigorik:
This PR  fixes the binding issue for the device-sided views that are passed into host-sided kernels.

---

TLDR: if the default space is device, but we call the host kernel - there will be an issue with argument memory layout (Left vs Right). This case should be handled by our side

---

When a _host-space kernel_ (e.g., pk.Serial) is run in a _device_ context, the Python side passes views that live in GPU memory (CudaUVMSpace, LayoutLeft), while **the generated C++ binding expects host views** with the host execution space’s layout (HostSpace, LayoutRight for Serial).

Pybind11 can't automatically cast these types, so we should handle that on our side.
This PR introduces function `_generate_mirror_with_exec_layout` which is handling this case

This PR also adds a compilation check. We should always check if the concrete module file exists, not the module directory.

### DIFF
--- a/pykokkos/core/compiler.py
+++ b/pykokkos/core/compiler.py
@@ -172,7 +172,7 @@ class Compiler:
         }:
             raise Exception(f"Types are required for style: {entity.style}")
 
-        if self.is_compiled(module_setup.output_dir):
+        if self.is_compiled(module_setup.output_dir) and module_setup.is_compiled():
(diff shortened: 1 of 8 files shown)


kennykos:
@IvanGrigorik Can you prove a code sample of what this PR is trying to fix? If so, can we integrate it into our unit tests?

PR #392, opened 2026-03-04
Remove vestigial code related to the depreciated workload kernel type

kennykos:


IvanGrigorik (commented):
Overall LG
Did you check this code on some nvidia gpu?

kennykos:
It seems like we are keeping workloads for now

IvanGrigorik:
> It seems like we are keeping workloads for now

- [ ] deprecate all examples
- [ ] deprecate benchmarks
- [ ] remove workload

list should look like that.

PR #397, opened 2026-03-18
MAINT: HIP build for AMD GPUs

kennykos:
Add `ENABLE_HIP` and `ARCH` CMake option, import Kokkos HIP spaces from the main namespace, and update documentation. With these additions, pykokkos runs on AMD GPUs with the HIP execution space!

Additionally, remove old CMake code needed to grab git submodules, which we do not use anymore.

Notes:
* To get pykokkos code to compile correctly, I needed to
```
export CXX=amdclang++
export CC=amdclang
```
* I build pykokkos with the following options
```
 python install_base.py install -- -DENABLE_LAYOUTS=ON -DENABLE_MEMORY_TRAITS=OFF -DENABLE_VIEW_RANKS=4 -DENABLE_CUDA=OFF -DENABLE_HIP=ON -DENABLE_THREADS=OFF -DENABLE_OPENMP=OFF -DARCH=AMD_GFX942 -DCMAKE_CXX_COMPILER=amdclang++
```

### DIFF
--- a/base/include/traits.hpp
+++ b/base/include/traits.hpp
@@ -61,14 +61,14 @@ DISABLE_TYPE(type_list<>)
 /// theoretically just say that python only supports up to 3 dimensions and if
 /// higher than that, user bindings must convert View to DynRankView.
 ///
-VIEW_DATA_DIMS(1, T *)
-VIEW_DATA_DIMS(2, T **)
-VIEW_DATA_DIMS(3, T ***)
-VIEW_DATA_DIMS(4, T ****)
-VIEW_DATA_DIMS(5, T *****)
-VIEW_DATA_DIMS(6, T ******)
-VIEW_DATA_DIMS(7, T *******)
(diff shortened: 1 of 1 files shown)


base/cmake/Modules/KokkosPythonKokkos.cmake
```diff
@@ -96,10 +97,18 @@ IF(_INTERNAL_KOKKOS)
         FIND_PACKAGE(CUDA QUIET)
     ENDIF()
 
+    IF(NOT DEFINED Kokkos_ENABLE_HIP)
+        FIND_PACKAGE(HIP QUIET)
+    ENDIF()
```
JBludau: Not asking to change it here, but should we think about versions we want to support? The easy decision would be to just pick the same range as Kokkos core

JBludau (approved):
The changes look good. But could you edit the title and description to be more descriptive? This pr also moves `HIP` out of experimental

PR #401, opened 2026-03-22
views: fix finalization for subview

IvanGrigorik:
https://github.com/kokkos/pykokkos/issues/321 added view registry and weakrefs to catch `finalize`-related issues (finalize after program termination), but there was no added same view registry entry for subviews. 

The ExaMiniMD benchmark also requires this fix, since it uses subviews. We also need it in general for view correctness,

### DIFF
--- a/pykokkos/interface/views.py
+++ b/pykokkos/interface/views.py
@@ -628,6 +628,14 @@ class Subview(ViewType):
         self.ndim = self.data.ndim
         self.size = self.data.size
 
+        try:
+            from pykokkos import _view_registry
+
+            _view_registry.add(self)
+        except (ImportError, AttributeError):
+            pass
+
+
     def _create_slice(self, data_slice: Union[slice, Tuple]) -> List[Union[int, slice]]:
         """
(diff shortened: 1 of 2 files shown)


pykokkos/interface/views.py
```diff
@@ -628,6 +628,13 @@ def __init__(
         self.ndim = self.data.ndim
         self.size = self.data.size
 
+        try:
+            from pykokkos import _view_registry
+
+            _view_registry.add(self)
+        except (ImportError, AttributeError):
+            pass
```
kennykos: Just a question, why would get `ImportError` or `AttributeError` here?

IvanGrigorik: removed.

pykokkos/interface/views.py
```diff
@@ -628,6 +628,10 @@ def __init__(
         self.ndim = self.data.ndim
         self.size = self.data.size
 
+        from pykokkos import _view_registry
```
kennykos, line 631: Can you make this import top level? And we should use relative imports, not absolute imports, no?

PR #404, opened 2026-03-29
core: spaces fix

IvanGrigorik:
This PR fixes the execution spaces.
As we discussed it with @kennykos, the default workspace describes the memory layout we will use.
There might be some issues if we use device as a default exec space, but host as a memory space. This PR fixes this.


IvanGrigorik:
Not the best fix idea. We should upgrade to Kokkos 5.0 and fix it after that. it should be much easier on newer kokkos version.

PR #405, opened 2026-03-30
Fix atomic fetch op pattern matching in workunit visitor

Wilson1211:
the regex shall be atomic_fetch_* rather than atomic_*,
otherwise, the logic will make atomic_fetch_op to be true when using atomic_compare_exchange, then the if statement : "if not is_atomic_increment and is_atomic_fetch_op and len(args) != 3:" will evaluated to true, which causes the problem


IvanGrigorik (approved):
LG, Thanks

PR #406, opened 2026-03-31
privatize pk.array

### LINKED ISSUE #388: Set `pk.array` as private (i.e., `pk._array`)
We don't want users to convert python array api compatible objects to views by hand. `pk.array` does such a conversion, and can lead to errors if used incorrectly. I don't think we need to expose this level of control to the user.

nayyirahsan:
Closes #388.

Renames `pk.array` to `pk._array` as the internal conversion function. `pk.array` is kept temporarily as a compatibility wrapper that emits a `DeprecationWarning` and forwards to `pk._array`. User-facing call sites are migrated to `pk.asarray`, while CuPy/GPU paths use `pk._array` directly to preserve conversion semantics. Deprecation and compatibility tests are added covering NumPy arrays, lists, scalars, and CuPy (CUDA-gated).

### DIFF
--- a/pykokkos/interface/views.py
+++ b/pykokkos/interface/views.py
@@ -5,6 +5,7 @@ import math
 from enum import Enum
 import os
 import sys
+import warnings
 from types import ModuleType
 from typing import Dict, Generic, Iterator, List, Optional, Tuple, TypeVar, Union
 
@@ -945,7 +946,7 @@ def is_array(array) -> bool:
     return True
 
 
-def array(
+def _array(
     array, space: Optional[MemorySpace] = None, layout: Optional[Layout] = None
 ) -> ViewType:
     """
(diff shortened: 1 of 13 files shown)


pykokkos/interface/views.py
```diff
@@ -971,6 +972,25 @@ def array(
...
+
+    warnings.warn(
+        "pk.array is deprecated and will be removed in a future release. "
+        "Use pk.asarray for user conversions; pk._array is internal/private.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+
+    return _array(array, space, layout)
+
```
IvanGrigorik, line 992: We can remove it completely.
No need to give warnings.

pykokkos/interface/views.py
```diff
@@ -980,7 +1000,17 @@ def asarray(obj, /, *, dtype=None, device=None, copy=None):
...
+    if np.isscalar(obj):
+        if obj in (pk.e, pk.pi, pk.inf):
+            is_array_api_constant = True
+        else:
+            try:
+                is_array_api_constant = bool(np.isnan(obj))
+            except TypeError:
+                is_array_api_constant = False
+
+    if is_array_api_constant:
```
IvanGrigorik, line 1013: What's the point of it? Why we can't use `pk.nan`?
And it looks invalid since we can test `cupy` arrays with `np.isnan`, which should give errors.

tests/test_views.py
```diff
@@ -97,9 +97,9 @@ def __init__(self, threads: int, i_1: int, i_2: int, i_3: int, i_4: int):
         cp_arr = cp.zeros((threads, 2)).astype(np.int32)
         list_arr = [np.array([0, 0], dtype=np.int32)] * threads
 
-        self.np_view: pk.View2D[int] = pk.array(np_arr)
-        self.cp_view: pk.View2D[int] = pk.array(cp_arr)
-        self.list_view: pk.View2D[int] = pk.array(list_arr)
+        self.np_view: pk.View2D[int] = pk.asarray(np_arr)
+        self.cp_view: pk.View2D[int] = pk._array(cp_arr)
+        self.list_view: pk.View2D[int] = pk.asarray(list_arr)
```
IvanGrigorik, line 102: The end goal is not to use `pykokkos` internal data structures. If we will deprecate `array` but still will use `asarray` - that's not correct. 
We want to have something like `self.np_view = np.array(...)` and be able to use it like that.
To be 100% flexible, we can have some intermediate data structure (per-file only) like `xp` that should decide if we want to use `cupy` or `numpy` based on current execution space.
But there should not be `xp.asarray(numpy_array)`. It should be `xp.array(...)`.

You can treat `xp` as a C++ macros like this:
```cpp
#ifdef __CUDA_ARCH__
    xp = cupy
#else
    xp = numpy
#endif
```

tests/test_views.py
```diff
@@ -376,6 +376,41 @@ def test_asarray_consts_vs_numpy(const, np_dtype, pk_dtype):
...
+    ],
+)
+def test_private_array_conversion_matches_asarray(arr):
+    from_private = pk._array(arr)
+    from_public = pk.asarray(arr)
+    assert type(from_private) is type(from_public)
+    assert_allclose(from_private, from_public)
+
+
+@pytest.mark.skipif(not HAS_CUDA, reason="CUDA/cupy not available")
```
IvanGrigorik, line 407: This one is cool. We should add more of this + more CUDA test.

IvanGrigorik (changes requested):

PR #412, opened 2026-04-22
Update install instructions for macOS

### LINKED ISSUE #390: macOS OpenMP build failure
- On macOS on Apple Silicon
- Homebrew OpenMP
- Built PyKokkos from source, main branch

When `@pk.workunit` is invoked, the JIT compiler calls `find_package(Kokkos)` which loads `KokkosConfig.cmake` which then calls
```
FIND_DEPENDENCY(OpenMP REQUIRED COMPONENTS CXX)
```
Apple Clang/CMake cannot find OpenMP though and so fails with
```
Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)
```

User must therefore set `CPATH` and `LIBRARY_PATH`. My current hack is to activate my `uv` venv with
```sh
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"

if [ ! -d "$SCRIPT_DIR/.venv" ]; then
    echo "Error: .venv not found. Run 'make setup' first."
    return 1 2>/dev/null || exit 1
fi

source "$SCRIPT_DIR/.venv/bin/activate"

if [ "$(uname)" = "Darwin" ]; then
    OMP_PREFIX="$(brew --prefix libomp 2>/dev/null)"
    if [ -n "$OMP_PREFIX" ]; then
        export CPATH="${OMP_PREFIX}/include${CPATH:+:$CPATH}"
        export LIBRARY_PATH="${OMP_PREFIX}/lib${LIBRARY_PATH:+:$LIBRARY_PATH}"
    fi
fi
```

kennykos:
Resolves #390.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,34 @@ PyKokkos consists of two components that need to be installed separately:
 
 #### Installing pykokkos-base
 
+<details>
+
+<summary>Known issue: macOS + homebrew OpenMP</summary>
+Homebrew does not add OpenMP to standard paths after
+an install, leading to CMake build failures of the form:
+
+```
+...
+Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)
+...
+```
+
+Such failures can be resolved by adding OpenMP to the `CPATH` and `LIBRARYPATH`:
+```
+export OMP_PREFIX=$(brew --prefix libomp)
+export CPATH="${OMP_PREFIX}/include${CPATH:+:$CPATH}"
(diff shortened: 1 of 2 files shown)


JBludau:
I am not against giving users hints but would refrain from giving recommendations.
I suspect homebrew has reasons for not adding openMP to the path per default and I suspect you use it together with Appleclang that comes on your macbook?

IvanGrigorik:
@kennykos, can we make a `Known issues` section? 
It would be easier to pile up everything that can't be fixed due to third-party dependency.

kennykos:
> @kennykos, can we make a `Known issues` section? It would be easier to pile up everything that can't be fixed due to third-party dependency.

Good idea, I will close this issue and open a `Known issues` PR.

PR #414, opened 2026-04-27
Test and support Kokkos::rsqrt 

kennykos:


pykokkos/core/translators/static.py
```diff
@@ -434,9 +434,8 @@ def translate_workunits(
                     workunit: cppast.MethodDecl = workunits[n][1]
                     self.add_rand_pool_state(workunit)
                     node_visitor.has_rand_call = False
-            except:
-                print(f"Translation of {w} {w.name} failed")
-                sys.exit(1)
```
JBludau, line 439: Hmm ... isn't that orthogonal? And is there a way to recover from this exception or should we basically abort?

kennykos: It's parallel to what we had before (i.e., `raise Exception` and  `sys.exit(1)` both abort the execution). The error was popping up on my end when developing this branch and I added the pythonic `Exception` for easier-to-parse tracebacks. 

The error is not fixable, but the reason we abort at this level and not lower is so we have access to the workunit name.

JBludau: I meant orthogonal to the change needed for using `rsqrt`. Meaning, this could be factored into a different PR and there we fix it in all places in the code if we want to

kennykos: Sounds good, I will revert the irrelevant changes.

JBludau (approved):

PR #415, opened 2026-04-29
Recompile stale `BASE_DIR`

### LINKED ISSUE #100: ENH, DOC: recompile on change
I'm not sure if we have this feature (we should document it if so), but it would be useful if instead of needing to blow away the `pk_cpp/` compiled asset folder when doing development work, `pykokkos` could accurately determine which kernels need to be recompiled based on changes in the Python source code. This does border on reinventing build system dependency graph logic I suppose. Could check what some other JIT/Python projects do in this space I suppose.

kennykos:
Closes #100, #234.

## Features

**1. AST-hashed compilation directories**
`pk_cpp` output directories now include an `AST_<hash>` subdirectory that encodes a structural hash of the compiled workunit. On each invocation, PyKokkos computes the current AST hash and compares it against the on-disk directory name; if the AST hashes differ, recompilation is triggered. This replaces the previous approach which **did not** check if the compiled AST matched the current AST.

**2. Incremental recompilation on AST change**
When a workunit's AST hash changes, `initialize_directory` renames the old `AST_<hash>` directory to the new hash rather than discarding it. Only `CMakeCache.txt` and `cmake.check_cache` are deleted (as they contain hardcoded absolute paths that CMake rejects after a rename), while all object files and link artifacts in `build/` are preserved. This allows CMake to perform an incremental rebuild rather than compiling from scratch.

**3. Recompilation unit test**
Adds `test_recompilation` which verifies end-to-end that modifying a `@pk.workunit` body triggers a correct recompile. The test writes a buggy kernel (`arr[i] += 2`), runs it and asserts the output, then overwrites it with the fixed kernel (`arr[i] += 1`) and asserts the recompiled output, confirming that stale builds are invalidated and that the new kernel produces correct results.

### DIFF
--- a/pykokkos/core/cpp_setup.py
+++ b/pykokkos/core/cpp_setup.py
@@ -125,11 +125,6 @@ class CppSetup:
         :param name: the name of the directory
         """
 
-        try:
-            shutil.rmtree(name)
-        except OSError:
-            pass
-
         # make the parent directory if necessary
         os.makedirs(name.parent, exist_ok=True)
 
--- a/pykokkos/core/cpp_setup.py
+++ b/pykokkos/core/cpp_setup.py
@@ -117,8 +117,10 @@ class CppSetup:
 
     def initialize_directory(self, name: Path) -> None:
         """
-        Creates an output directory, overwriting an existing directory with the same name.
-        Checks if an older AST directory exists, and if so copies over the relevant information.
+        Checks if an older AST directory exists;
+        if so, copies over the relevant information,
+        otherwise, creates an output directory,
+        overwriting an existing directory with the same name.
 
         :param name: the name of the directory
         """
@@ -129,16 +131,13 @@ class CppSetup:
             pass
 
         # make the parent directory if necessary
-        try:
-            os.makedirs(name.parent, exist_ok=True)
-        except FileExistsError:
-            pass
+        os.makedirs(name.parent, exist_ok=True)
 
         # check if an older AST hash exists
         old_dir = [
             f for f in name.parent.parent.iterdir() if f.is_dir() and f != name.parent
-        ]
-        if len(old_dir) == 1:
+        ]  # this is expected to be 0 or 1 entries
+        if old_dir:
             # if an older AST directory exists, change the name to the new AST directory
             old_dir = old_dir[0]
             old_dir.rename(name.parent)
@@ -153,10 +152,7 @@ class CppSetup:
                 stale_cache_check.unlink()
 
         # make the name directory if necessary
-        try:
-            os.makedirs(name, exist_ok=True)
-        except FileExistsError:
-            pass
+        os.makedirs(name, exist_ok=True)
 
     def write_source(
         self,
--- a/pykokkos/core/cpp_setup.py
+++ b/pykokkos/core/cpp_setup.py
@@ -117,7 +117,8 @@ class CppSetup:
 
     def initialize_directory(self, name: Path) -> None:
         """
-        Creates an output directory, overwriting an existing directory with the same name
+        Creates an output directory, overwriting an existing directory with the same name.
+        Checks if an older AST directory exists, and if so copies over the relevant information.
 
         :param name: the name of the directory
         """
@@ -127,6 +128,31 @@ class CppSetup:
         except OSError:
             pass
 
+        # make the parent directory if necessary
+        try:
+            os.makedirs(name.parent, exist_ok=True)
+        except FileExistsError:
+            pass
+
+        # check if an older AST library exists
+        old_dir = [
+            f for f in name.parent.parent.iterdir() if f.is_dir() and f != name.parent
+        ]
+        if len(old_dir) == 1:
+            # if an older AST directory exists, change the name to the new AST directory
+            old_dir = old_dir[0]
+            old_dir.rename(name.parent)
+            # CMakeCache.txt has hardcoded absolute paths to the old AST directory.
+            # Delete only the cache files so CMake reconfigures with correct paths
+            # while keeping object files in build/ for incremental recompilation.
+            stale_cache = name / "build" / "CMakeCache.txt"
+            if stale_cache.is_file():
+                stale_cache.unlink()
+            stale_cache_check = name / "build" / "CMakeFiles" / "cmake.check_cache"
+            if stale_cache_check.is_file():
+                stale_cache_check.unlink()
+
+        # make the name directory if necessary
         try:
             os.makedirs(name, exist_ok=True)
         except FileExistsError:
--- a/pykokkos/core/cpp_setup.py
+++ b/pykokkos/core/cpp_setup.py
@@ -134,7 +134,7 @@ class CppSetup:
         except FileExistsError:
             pass
 
-        # check if an older AST library exists
+        # check if an older AST hash exists
         old_dir = [
             f for f in name.parent.parent.iterdir() if f.is_dir() and f != name.parent
         ]
--- a/pykokkos/core/module_setup.py
+++ b/pykokkos/core/module_setup.py
@@ -169,7 +169,20 @@ class ModuleSetup:
         restrict_signature: Optional[str] = None,
     ) -> Optional[Path]:
         """
-        Get the output directory for an execution space
+        Get the output directory for an execution space.
+
+        The directory structure is hierarchical to allow for efficient caching.
+        The AST signature is given highest priority as it defines the
+        core logic of the kernel; a change in the AST necessitates a new
+        compilation regardless of other parameters. The type signature
+        and restrict signature are used to avoid excessive recompilation;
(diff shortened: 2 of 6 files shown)


pykokkos/core/parsers/parser.py
```diff
@@ -79,6 +80,10 @@ def __init__(self, path: Optional[str], pk_import: Optional[str] = None):
             self.functors: Dict[str, PyKokkosEntity] = {}
             self.workunits: Dict[str, PyKokkosEntity] = {}
 
+        # get parser signature
+        signature = ast.dump(self.tree)
+        self.signature = hashlib.md5(signature.encode()).hexdigest()
```
gliga, line 85: how expensive is this? I assume our ast's for kernels are small that this should not be an issue?

pykokkos/core/cpp_setup.py
```diff
@@ -127,6 +128,31 @@ def initialize_directory(self, name: Path) -> None:
         except OSError:
             pass
 
+        # make the parent directory if necessary
+        try:
+            os.makedirs(name.parent, exist_ok=True)
+        except FileExistsError:
+            pass
+
+        # check if an older AST library exists
```
gliga: AST library? you mean AST hash?

tests/test_recompile.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # load buggy source
+    mod_buggy = _load_fresh(kernel_file)
+
+    # run the buggy kernel
+    n = 5
+    arr_buggy = np.zeros(n, dtype=np.int32)
+    pk.parallel_for(n, mod_buggy.add1, arr=arr_buggy)
+
+    # assert buggy array is correct
```
gliga: sounds weird

tests/test_recompile.py
```diff
@@ -0,0 +1,98 @@
...
+
+    arr_correct = np.zeros(n, dtype=np.int32)
+    pk.parallel_for(n, mod_correct.add1, arr=arr_correct)
+    expected = np.zeros(n, dtype=np.int32) + 1
+    try:
+        np.testing.assert_equal(arr_correct, expected)
+    except AssertionError as e:
+        raise AssertionError(
+            f"kernel is incorrect\nactual:  {arr_correct}\ndesired: {expected}"
+        ) from e
```
gliga, line 98: this test does not check that values of signature on disk changed or something was renamed

gliga (approved):
Overall, good change. See comments.

gliga:
I do not like the renaming part, but we can keep it and see when/if it breaks.

gliga:
Are we ever going to change the name of that directory? It has been a while that we wanted to move to `.pykokkos` or we can use `.pk`

pykokkos/core/parsers/parser.py
kennykos: I didn't do a performance test. This block is only done when the workunit is first seen by pykokkos, so it is a startup cost that will not add overhead during repeated execution. 

In most cases the ast's should be relatively small; e.g., in my Ewald code, the slowest time for this block of code is `0.013` seconds.

tests/test_recompile.py
kennykos: I kept it abstract in case we want to change in implementation details down the road (i.e., if we want to move away from renaming or change the directory structure).

kennykos:
> Are we ever going to change the name of that directory? It has been a while that we wanted to move to `.pykokkos` or we can use `.pk`

Changed to `.pykokkos` in #416.

pykokkos/core/cpp_setup.py
```diff
@@ -117,7 +117,8 @@ def compile(
 
     def initialize_directory(self, name: Path) -> None:
         """
-        Creates an output directory, overwriting an existing directory with the same name
+        Creates an output directory, overwriting an existing directory with the same name.
+        Checks if an older AST directory exists, and if so copies over the relevant information.
```
JBludau: maybe we want to reverse the order of the comments?

pykokkos/core/cpp_setup.py
```diff
@@ -127,6 +128,31 @@ def initialize_directory(self, name: Path) -> None:
         except OSError:
             pass
 
+        # make the parent directory if necessary
+        try:
+            os.makedirs(name.parent, exist_ok=True)
+        except FileExistsError:
+            pass
+
```
JBludau, line 130: why would we want to catch a FileExistsError if we pass `exists_ok=true`?

pykokkos/core/cpp_setup.py
```diff
@@ -127,6 +128,31 @@ def initialize_directory(self, name: Path) -> None:
...
+        # check if an older AST hash exists
+        old_dir = [
+            f for f in name.parent.parent.iterdir() if f.is_dir() and f != name.parent
+        ]
+        if len(old_dir) == 1:
```
JBludau: why a list if we just expect 0 or 1 entries?

pykokkos/core/module_setup.py
```diff
@@ -180,6 +192,8 @@ def get_output_dir(
             out_dir = out_dir / f"types_{types_signature}"
         if restrict_signature is not None:
             out_dir = out_dir / f"restrict_{restrict_signature}"
+        if ast_signature is not None:
+            out_dir = out_dir / f"AST_{ast_signature}"
```
JBludau, line 209: hmm so we have a hierarchy which has the ast signature as the highest priority ... but is there cases where the ast signature would not trigger but the others would?

tests/test_recompile.py
```diff
@@ -0,0 +1,98 @@
...
+    # run the buggy kernel
+    n = 5
+    arr_buggy = np.zeros(n, dtype=np.int32)
+    pk.parallel_for(n, mod_buggy.add1, arr=arr_buggy)
+
+    # assert add2 array is correct
+    try:
+        np.testing.assert_equal(arr_buggy, np.zeros(n, dtype=np.int32) + 2)
+    except AssertionError as e:
+        raise AssertionError("buggy kernel is incorrect") from e
```
JBludau: maybe not call it `buggy` but rather v1 and v2?

pykokkos/core/module_setup.py
kennykos: Yes, consider the scenario:

* workunit v0 is compiled with pk.float arrays;
* an edit is made and we now have ATS has for workunit v1;
* workunit v1 is compiled with pk.double arrays;
* workunit v1 is called again with pk.float arrays. In this case, the type hash would trigger, but the ats hash would not as it is out of date.

pykokkos/core/cpp_setup.py
kennykos: I chose list comprehension as it seemed the most pythonic thing to do, added a comment to specify that we have 0 or 1 entries.

pykokkos/core/cpp_setup.py
kennykos: Removed in [ee7e739](https://github.com/kokkos/pykokkos/pull/415/commits/ee7e739fd80d2799a5dfed21071de69609a4d76f), this was a vestigial code pattern.

pykokkos/core/cpp_setup.py
kennykos: Changed in [ee7e739](https://github.com/kokkos/pykokkos/pull/415/commits/ee7e739fd80d2799a5dfed21071de69609a4d76f).

tests/test_recompile.py
kennykos: Addressed in [19b56c0](https://github.com/kokkos/pykokkos/pull/415/commits/19b56c03bc2f00d56db9d851d176beef05c8eb57).

kennykos:
@JBludau I want to make sure you're aware of the new commit [a810722](https://github.com/kokkos/pykokkos/pull/415/commits/a810722b97ebb68b72cd9a369a11cad912ebb519).

pykokkos/core/module_setup.py
JBludau: but doesn't the AST change if we change types from float to double? Even unmanaged Views change their type, I am surprised that this would not trigger an AST hash fail

kennykos: Yes, the AST will change, but we want to keep both float and double types cached in this scenario to avoid excessive recompilation (e.g., an adaptive time-stepper may use a float and double execution to get error estimates)

JBludau: I see. Could you add a comment explaining this intent? Future selves will thank us for it

JBludau (approved):
@kennykos I think the hierarchy of rebuilds could have a comment to express the intent, but non blocking.

pykokkos/core/module_setup.py
kennykos: Addressed in a3a23c8597f85d4c7327d1cd4ccf9194adb0e76f

PR #418, opened 2026-05-06
set UVM off by default if CUDA is enabled (deprecated by Kokkos)

IvanGrigorik:
This PR removes the deprecated CUDA `UVM` memory space option, which causes slowdowns due to page-fetch faults during kernel execution.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -110,6 +110,8 @@ install ``base`` with required CMake flags (example performs an install with  Op
      - Build Kokkos submodule with HIP support
    * - ARCH
      - `Kokkos GPU architecture <https://kokkos.org/kokkos-core-wiki/API/core/Macros.html#architectures>`_
+   * - Kokkos_ENABLE_CUDA_UVM
+     - Build Kokkos submodule with CUDA UVM support. Off by default.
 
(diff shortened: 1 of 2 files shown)


JBludau (commented):
Since the option will be removed at some point, why not just delete it completely? The default for that option is OFF anyways

kennykos:
Also, this change should be reflected in the CMake flags section of the docs.

docs/installation.rst
```diff
@@ -110,6 +110,8 @@ install ``base`` with required CMake flags (example performs an install with  Op
      - Build Kokkos submodule with HIP support
    * - ARCH
      - `Kokkos GPU architecture <https://kokkos.org/kokkos-core-wiki/API/core/Macros.html#architectures>`_
+   * - Kokkos_ENABLE_CUDA_UVM
+     - Build Kokkos submodule with CUDA UVM support. Off by default.
```
JBludau, line 114: do we still need this then?

IvanGrigorik: People might need to turn it on
We support it, so I see no reason to not add it into documentation

JBludau (approved):
ok

PR #419, opened 2026-05-13
Test fmin and fmax functions

kennykos:


tests/test_ufuncs.py
```diff
@@ -101,6 +101,18 @@ def rsqrt_workunit(tid: int, view: pk.View1D[pk.double]) -> None:
     view[tid] = rsqrt(view[tid])
 
 
+@pk.workunit
+def fmin_workunit(tid: int, view: pk.View1D[pk.double]) -> None:
+    nan: pk.double = 1 / 0 - 1 / 0
```
JBludau, line 106: hmm ... maybe testing against nan is not the best idea since it compares not equal to itself. 
How about using the max/min val for double?

IvanGrigorik (changes requested):
Please use FMAX / FMIN to test it

kennykos:
> Please use FMAX / FMIN to test it

what do you mean by this?

PR #421, opened 2026-05-25
feat: Fix Reducer Mismatch

WeiTaShao:
## Summary

This PR fixes a reducer type mismatch issue.

## Changes

- Fix reducer type handling
- Ensure reducer-related logic uses the expected type
- Improve compatibility with the existing parallel_reduce behavior

## Testing

- Modified tests/test_parallelreduce_mod.py
- Self generated script: check_reducer.py


kennykos:
The format run is failing. Can you make sure you are using `black==25.12.0`? Thanks.

activate_pyk.sh
```diff
@@ -0,0 +1,11 @@
...
+source ~/miniconda3/etc/profile.d/conda.sh
+conda activate pyk
+
+# Force Kokkos + nvcc to use GCC 13 (CUDA 12.4 only supports GCC <= 13)
+export CC=/usr/bin/gcc-13
+export CXX=/usr/bin/g++-13
+export CUDAHOSTCXX=/usr/bin/g++-13
+
+# Force nvcc_wrapper to use g++-13 instead of /usr/bin/c++
+export NVCC_WRAPPER_DEFAULT_COMPILER=/usr/bin/g++-13
```
IvanGrigorik: remove or change system interoperability. Not everyone has g++13 (or conda / cuda)

check_reducer.py
IvanGrigorik, line 1: I suppose this file should not exist or should be inside the `tests` directory.
Remove or move to `tests` + make pytest-compatible

pykokkos/core/translators/bindings.py
```diff
@@ -10,6 +10,51 @@
...
+MINMAX_LOC_REDUCERS = {"MinMaxLoc"}
+NON_SCALAR_REDUCERS = VALUE_LOC_REDUCERS | MINMAX_REDUCERS | MINMAX_LOC_REDUCERS
+
+
+def is_non_scalar_reducer(reducer: Optional[str]) -> bool:
+    return reducer in NON_SCALAR_REDUCERS
+
+
+def get_reducer_scalar_type(acc_type: str) -> str:
+    match = re.match(r"Kokkos::(?:MaxLoc|MinLoc|MinMax|MinMaxLoc)<([^,>]+)", acc_type)
```
IvanGrigorik: Cool, I love regex.
However, in the future, try not to use them.
Recently, we found that we have some unhandled edge cases related to regex https://github.com/kokkos/pykokkos/pull/405, and if we will have more regexes, we have to make sure that all of them are handled (which would be a pretty hard engineering effort)

pykokkos/core/type_inference/args_type_inference.py
```diff
@@ -478,7 +478,10 @@ def get_type_info(
 
     is_standalone_workunit: bool = True
     list_passed: bool = True
-    execution_space: ExecutionSpace = policy.space.space
+    policy_space = policy.space
+    execution_space: ExecutionSpace = (
+        policy_space.space if hasattr(policy_space, "space") else policy_space
+    )
```
IvanGrigorik: Why do we need to do that? What was wrong with the previous implementation?

pykokkos/core/visitors/workunit_visitor.py
```diff
@@ -10,6 +10,34 @@
 from .pykokkos_visitor import PyKokkosVisitor
 
 
+VALUE_LOC_REDUCERS = {"MaxLoc", "MinLoc"}
+MINMAX_REDUCERS = {"MinMax"}
+MINMAX_LOC_REDUCERS = {"MinMaxLoc"}
+NON_SCALAR_REDUCERS = VALUE_LOC_REDUCERS | MINMAX_REDUCERS | MINMAX_LOC_REDUCERS
```
IvanGrigorik: Why do we need to define it again? Just use it from `bindings.py` if there is no circular import error.

pykokkos/core/runtime.py
```diff
@@ -37,13 +37,26 @@
...
 from .compiler import Compiler
 from .module_setup import EntityMetadata, get_metadata, ModuleSetup
 from .run_debug import run_workload_debug, run_workunit_debug
 
 
+def reducer_cache_signature(reducer: Optional[Reducer]) -> Optional[str]:
+    if reducer is None:
+        return None
+
+    return reducer.name
```
IvanGrigorik, line 52: Cache based only on name?

 If so, we need to change it (at least get the hash of the workunit AST)
I know, that might sound like a quibble, so we can do it in the follow-up PR. It is something that we have wanted to fix in PyKokkos for a long time.

pykokkos/core/runtime.py
```diff
@@ -832,10 +852,11 @@ def get_module_setup(
         self,
         entity: Union[object, Callable[..., None]],
         space: ExecutionSpace,
-        ast_signature: str,
+        ast_signature: Optional[str] = None,
```
IvanGrigorik: AST signature must be required for module setup. We compute the module hash and do (re-)compilation based on this information.

pykokkos/interface/__init__.py
```diff
@@ -77,6 +77,21 @@
...
+    MaxLoc,
+    Min,
+    MinLoc,
+    MinMax,
+    MinMaxLoc,
+    Prod,
+    Reducer,
+    Sum,
+)
 from .random import rand, RandomPool, Random_XorShift64_Pool, Random_XorShift1024_Pool
```
IvanGrigorik, line 95: Cool, thanks

IvanGrigorik (changes requested):
This PR is ok.
However, two main issues:
1) A lot of changes
2) A lot of **unexplained** changes. 

I assume to address GitHub comments and insert some explanatory comments in the future (especially when we change something that is already working). 
Also, please append your PR main comment. If we need to fix something, give an example of why we need to fix it (how it works now vs how it should work). Like https://github.com/kokkos/pykokkos/pull/381. Or this, when you are addressing some existing issue: https://github.com/kokkos/pykokkos/pull/344

---

And, as Gabriel said, use `black==25.12.0` to address formatting issues.

gliga:
The key is that we want each PR to remain small (unless absolutely necessary to blend many changes together).

gliga:
> The format run is failing. Can you make sure you are using black==25.12.0? Thanks.
Why are we not enforcing this automatically? @kennykos @IvanGrigorik

check_reducer.py
gliga: I agree we already have tests you added, why another file?

IvanGrigorik:
> > The format run is failing. Can you make sure you are using black==25.12.0? Thanks.
> > Why are we not enforcing this automatically? @kennykos @IvanGrigorik

I suppose that we can add the `dev` branch and have quite a different `conda-env` file in it. Currently, there is no `black` in the conda file (which makes sense, since we don't need it by default).
We can set up something like a `.clang-format` file, which will force the formatter to change files in a specific order, but if something is introduced in a future version and we don't specify that explicitly, we will have the same issue.

Should we just put `black==25.12.0` in the conda env file by default?

kennykos:
> > > The format run is failing. Can you make sure you are using black==25.12.0? Thanks.
> > > Why are we not enforcing this automatically? @kennykos @IvanGrigorik
> 
> I suppose that we can add the `dev` branch and have quite a different `conda-env` file in it. Currently, there is no `black` in the conda file (which makes sense, since we don't need it by default).
> We can set up something like a `.clang-format` file, which will force the formatter to change files in a specific order, but if something is introduced in a future version and we don't specify that explicitly, we will have the same issue.
> 
> Should we just put `black==25.12.0` in the conda env file by default?

Why don't we have a GitHub action that auto-formats the code on a push?

gliga:
> > > > The format run is failing. Can you make sure you are using black==25.12.0? Thanks.
> > > > Why are we not enforcing this automatically? @kennykos @IvanGrigorik
> > 
> > 
> > I suppose that we can add the `dev` branch and have quite a different `conda-env` file in it. Currently, there is no `black` in the conda file (which makes sense, since we don't need it by default).
> > We can set up something like a `.clang-format` file, which will force the formatter to change files in a specific order, but if something is introduced in a future version and we don't specify that explicitly, we will have the same issue.
> > Should we just put `black==25.12.0` in the conda env file by default?
> 
> Why don't we have a GitHub action that auto-formats the code on a push?

I like adding black to conda env file by default

pykokkos/core/type_inference/args_type_inference.py
WeiTaShao: The previous implementation assumed policy.space is always an ExecutionSpaceInstance and therefore has a .space field. That is true for most RangePolicy/TeamPolicy cases, but not for policies that keep a raw ExecutionSpace, such as MDRangePolicy and Debug. In those cases policy.space.space raises AttributeError.

pykokkos/core/visitors/workunit_visitor.py
WeiTaShao: Importing these from bindings.py creates a circular import because `bindings.py` imports pykokkos.core.visitors, which imports workunit_visitor. I will move the shared reducer helpers into a small neutral utility module and import it from both `bindings.py` and `workunit_visitor.py`.

check_reducer.py
WeiTaShao: The original test file is `tests/test_parallelreduce.py`. However, this test file only tests the single reducer Sum, so I created another test file called `tests/test_parallelreduce_mod.py`. Do I need to replace the old file with the modified one, or can I leave them as two separate files?

activate_pyk.sh
WeiTaShao: Removed

pykokkos/core/runtime.py
WeiTaShao: Fixed

check_reducer.py
kennykos: You should move all your tests into the `tests/test_parallelreduce.py` file.

pykokkos/core/type_inference/args_type_inference.py
kennykos: Great catch. This should be it's own issue, all spaces should have the same attribute structure (and we should test to ensure this).

check_reducer.py
WeiTaShao: Moved

IvanGrigorik:
Something wrong with ARM workflow. It’s not related to your changes

pykokkos/core/compiler.py
```diff
@@ -134,6 +134,7 @@ def compile_object(
         updated_types: Optional[UpdatedTypes],
         types_signature: Optional[str],
         restrict_views: Set[str],
+        reducer: Optional[str] = None,
```
IvanGrigorik: Can we put it in `module_setup` instead of yet another argument?
Like make `reducer` as a workunit module metadata and pass this metadata to the  static translator here
https://github.com/WeiTaShao/pykokkos/blob/13b9877ea14395cabff2ab6854f7d3c9a887c982/pykokkos/core/compiler.py#L257-L262
Since we insert `reducer_name` in module setup, it won't hurt us.

pykokkos/core/runtime.py
```diff
@@ -288,7 +308,7 @@ def run_workunit(
             return future
 
         return self.execute_workunit(
-            name, policy, workunit, operation, parser, **kwargs
+            name, policy, workunit, operation, parser, reducer=reducer, **kwargs
         )
```
IvanGrigorik, line 309: Please pass it via `**kwargs` (like add reducer to **kwargs for `run_workunit`. 
This way, you don't need to modify each LOC just to pass yet another argument to a new function. 
Just add to `**kwargs` and parse it at the endpoint of interest (in our case, it is  `precompile_workunit` and `get_module_setup` functions)
This operation should be generic enough.

IvanGrigorik (commented):
Overall LG, couple of comments

kennykos:
Looks good. One more request (sorry): can you add a simple example of each reducer type (i.e.,     MINMAX_LOC_REDUCERS, MINMAX_REDUCERS, VALUE_LOC_REDUCERS)  to the `https://github.com/kokkos/pykokkos/tree/main/examples/pykokkos` directory?

examples/pykokkos/reducers.py
```diff
@@ -0,0 +1,156 @@
+import numpy as np
```
kennykos, line 1: Looks great! Thanks

pykokkos/core/translators/bindings.py
```diff
@@ -273,7 +311,8 @@ def get_return_type(operation: str, workunit: cppast.MethodDecl) -> str:
     if acc_decl is None:
         return_type = "void"
     else:
-        return_type = acc_decl.decltype.typename.value
+        typename = acc_decl.decltype.typename
+        return_type = typename.value if hasattr(typename, "value") else typename
```
kennykos, line 321: Will this be fixed by #423 or is it something else?

pykokkos/core/translators/bindings.py
```diff
@@ -445,20 +492,32 @@ def generate_kernel(
...
+    return_type: str = "pybind11::tuple" if is_non_scalar_reducer(reducer) else acc_type
     signature: str = generate_kernel_signature(return_type, kernel, params)
 
     acc: str = ""
     if operation in ("reduce", "scan"):
-        acc = f"{return_type} {Keywords.Accumulator.value} = 0;"
+        if operation == "reduce" and reducer is not None:
+            acc = f"{acc_type} {Keywords.Accumulator.value};"
+        else:
+            acc = f"{acc_type} {Keywords.Accumulator.value} = 0;"
```
kennykos, line 510: Do we support a variable different initial value for reducers? E.g., for a product reduction, don't I want my initial value to be 1? If this isn't supported we should open a separate issue.

pykokkos/core/translators/reducer_util.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def is_non_scalar_reducer(reducer: Optional[str]) -> bool:
+    return reducer in NON_SCALAR_REDUCERS
+
+
+def get_cpp_type_name(decltype: cppast.Type) -> str:
+    if isinstance(decltype, cppast.PrimitiveType):
+        typename = decltype.typename
+        return typename.value if hasattr(typename, "value") else typename
```
kennykos, line 21: Again, is this branching resolved by #423?

IvanGrigorik (approved):
TODO in the follow-up PR:
1) Cache based on workunit AST + reducer name, not on just the reducer name 
2) Fix execution spaces (it should always have the ability to do `policy_space.space` or just `policy_space`. Need to unify.

IvanGrigorik:
DO NOT merge till all tests are passed. I will also resolve this PR mismatch now. Gabriel did his PR later, but a lot of stuff from this PR can be simplified by his changes.

IvanGrigorik (changes requested):

kennykos:
@WeiTaShao #423 has been merged into main; can you merge it into this branch then clean up the `space.space` branching?

WeiTaShao:
> @WeiTaShao #423 has been merged into main; can you merge it into this branch then clean up the `space.space` branching?

Done

pykokkos/core/translators/bindings.py
```diff
@@ -8,6 +8,40 @@
...
+
+def get_reducer_expr(reducer: str, scalar_type: str) -> str:
+    if reducer in VALUE_LOC_REDUCERS:
+        return f"Kokkos::{reducer}<{scalar_type},int>({Keywords.Accumulator.value})"
+    if reducer in MINMAX_REDUCERS:
+        return f"Kokkos::{reducer}<{scalar_type}>({Keywords.Accumulator.value})"
+    if reducer in MINMAX_LOC_REDUCERS:
+        return f"Kokkos::{reducer}<{scalar_type},int>({Keywords.Accumulator.value})"
+
+    return f"Kokkos::{reducer}<{scalar_type}>({Keywords.Accumulator.value})"
```
kennykos: When does this statement return? Also, can we use `elif` for clarity?

pykokkos/core/translators/bindings.py
```diff
@@ -8,6 +8,40 @@
...
+        return f"pybind11::make_tuple({Keywords.Accumulator.value}.min_val,{Keywords.Accumulator.value}.max_val)"
+    if reducer in MINMAX_LOC_REDUCERS:
+        return (
+            f"pybind11::make_tuple({Keywords.Accumulator.value}.min_val,"
+            f"{Keywords.Accumulator.value}.min_loc,"
+            f"{Keywords.Accumulator.value}.max_val,"
+            f"{Keywords.Accumulator.value}.max_loc)"
+        )
+
+    return Keywords.Accumulator.value
```
kennykos: When does this statement return? Also, can we use `elif` for clarity?

pykokkos/core/translators/bindings.py
```diff
@@ -24,6 +58,10 @@ def is_hierarchical(workunit: Optional[cppast.MethodDecl]) -> bool:
     # Iterate over each parameter (skipping the tag)
     for p in workunit.params[1:]:
         if isinstance(p.decltype, cppast.ClassType):
+            if p.decltype.typename.startswith(
```
kennykos, line 67: Why do we need this statement? Can you add a comment?

pykokkos/core/translators/bindings.py
kennykos: @WeiTaShao I'm guessing we'll need a small PR to address this branching, similar to #423. Can you give a brief description of what is broken and I can fix it?

tests/test_parallelreduce.py
```diff
@@ -66,7 +189,7 @@ def test_squaresum_types(series_max, dtype):
     np_data = np.arange(series_max, dtype=dtype)
     expected = np.sum(np_data**2)
 
-    view = pk.array(np_data)
+    view = pk.array(np_data, space=pk.HostSpace)
```
kennykos, line 192: I didn't know we had `pk.HostSpace`. Cool!

pykokkos/core/translators/bindings.py
WeiTaShao: The issue is that `acc_decl.decltype.typename` is not structurally consistent.
For primitive accumulator types it can be an enum-like object with a .value
field, but for non-scalar reducer value_type it is already a string such as
`Kokkos::MaxLoc<double,int>::value_type`. The old code assumed .value always
exists, so reducer value_type return types raise AttributeError. I added the
branch only to handle both representations.

pykokkos/core/translators/bindings.py
WeiTaShao: This fallback is for the scalar built-in reducers (such as pk.Sum, pk.Prod) and yes using elif will be much more clear as the reducer categories are mutually exclusive

pykokkos/core/translators/bindings.py
WeiTaShao: This return is for normal scalar reduction and scalar built-in reducers. As non-scalar reducers will return tuple in the early branch

pykokkos/core/translators/bindings.py
WeiTaShao: Because a non-scalar reducer accumulator is represented as a Kokkos reducer `value_type`, it is also a `ClassType`. If we do not skip it, the bindings may incorrectly classify the reducer workunit as a hierarchical workunit. This is to prevent the reducer `value_type` from being mistakenly identified as a `TeamMember`.

pykokkos/core/runtime.py
```diff
@@ -37,13 +37,25 @@
...
 
 
+def reducer_cache_signature(reducer: Optional[Reducer]) -> Optional[str]:
+    if reducer is None:
+        return None
+
+    return reducer.name
+
+
+def get_policy_execution_space(policy: ExecutionPolicy) -> ExecutionSpace:
```
kennykos: We don't need this function now that #423 unified execution space.

pykokkos/core/translators/bindings.py
kennykos: Please change to 
```
if:
...
elif:
...
else:
raise ValueError(f"unrecognized reducer {reducer}")
```

The phantom return makes the code hard to understand.

pykokkos/core/translators/bindings.py
kennykos: Please change to 
```
if:
...
elif:
...
else:
raise ValueError(f"unrecognized reducer {reducer}")
```

The phantom return makes the code hard to understand.

pykokkos/interface/parallel_dispatch.py
```diff
@@ -37,6 +38,10 @@
 }
 
 
+def get_policy_execution_space(policy: ExecutionPolicy) -> ExecutionSpace:
```
kennykos: Again, we don't need this function given #423. Also, in the future, we should have only one version of a function and import it to different files, this reduces bloat in the codebase.

kennykos (changes requested):
Nearing the finish line, just need to remove some vestigial code and clear up branching returns.

pykokkos/interface/parallel_dispatch.py
WeiTaShao: Done

pykokkos/core/runtime.py
WeiTaShao: Fixed

pykokkos/core/translators/bindings.py
```diff
@@ -8,6 +8,45 @@
...
+    if reducer in VALUE_LOC_REDUCERS:
+        return f"Kokkos::{reducer}<{scalar_type},int>({Keywords.Accumulator.value})"
+    elif reducer in MINMAX_REDUCERS:
+        return f"Kokkos::{reducer}<{scalar_type}>({Keywords.Accumulator.value})"
+    elif reducer in MINMAX_LOC_REDUCERS:
+        return f"Kokkos::{reducer}<{scalar_type},int>({Keywords.Accumulator.value})"
+    elif reducer in SCALAR_REDUCERS:
+        return f"Kokkos::{reducer}<{scalar_type}>({Keywords.Accumulator.value})"
+
+    raise ValueError(f"unrecognized reducer {reducer}")
```
kennykos, line 31: Thanks!

kennykos (approved):
Great work! I'll merge in after the CI runs.

kennykos:
@IvanGrigorik waiting on your review; you requested changes.

IvanGrigorik (approved):
I'm not fan of such PRs but I can live with it for now

PR #422, opened 2026-05-28
Assert that numpy-like arrays are contiguous

kennykos:
This stems form an issue where a PyKokkos kernel was passed a non-contiguous numpy array, leading to a segfault. This should be caught on the PyKokkos side.

### DIFF
--- a/pykokkos/interface/views.py
+++ b/pykokkos/interface/views.py
@@ -964,6 +964,9 @@ def array(
     :returns: a PyKokkos View wrapping the array
     """
 
+    if not array.flags["F_CONTIGUOUS"] and not array.flags["C_CONTIGUOUS"]:
+        raise ValueError(f"numpy array is not contiguous")
+
     # if numpy array, use from_numpy()
     if isinstance(array, np.ndarray) or np.isscalar(array):
         return from_numpy(array, space, layout)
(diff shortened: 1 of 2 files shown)


pykokkos/interface/views.py
```diff
@@ -960,15 +964,27 @@ def array(
...
+    # if an array is not a recognized type, try coasting it to a numpy array
+    if (
+        not isinstance(array, np.ndarray)
+        and not np.isscalar(array)
+        and not is_array(array)
+    ):
+        array = np.asarray(array)
+
+    # check that the array is contiguous
+    if not array.flags["F_CONTIGUOUS"] and not array.flags["C_CONTIGUOUS"]:
```
IvanGrigorik, line 976: This will cause a failure if the array is `np.array`, not an `np.ndarray`, since simple arrays do not have a `flags` field.
AFAIK, we can pass `np.array`, so hasattr should be essential here.
do `if hasattr(array, flags) and ...`

IvanGrigorik (approved):
ok, one comment which confuses me a bit

pykokkos/interface/views.py
kennykos: I believe a numpy array is always type `numpy.ndarray` (see https://numpy.org/doc/stable/reference/arrays.ndarray.html). 

E.g., 

```Py
>>> import numpy as np
>>> arr = np.array([1, 2, 3])
>>> type(arr)
<class 'numpy.ndarray'>
```

IvanGrigorik: ok

PR #425, opened 2026-06-11
Add TranslationError class

kennykos:
This allows for Python tracebacks as opposed to using sys.exit

### DIFF
--- a/pykokkos/core/visitors/visitors_util.py
+++ b/pykokkos/core/visitors/visitors_util.py
@@ -176,6 +176,7 @@ class TranslationError(Exception):
             )
             msg += " " * self.node.col_offset + "^" * err_len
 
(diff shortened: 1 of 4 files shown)


pykokkos/core/visitors/visitors_util.py
```diff
@@ -10,8 +10,12 @@
 from pykokkos.interface import Layout, MemorySpace, Trait
 
 
+def pretty_format(node):
```
IvanGrigorik: Either specify the type for `node` or name the function `pretty_format_ast`.
The best thing is to do both.

IvanGrigorik (approved):
LG, one comment.
You can also try using the [`colorama`](https://pypi.org/project/colorama/) library instead of ANSI escape codes, but it's up to you; I will not block on this.

PR #428, opened 2026-06-26
core: update mypy and resolve issues

kennykos:
The CI was using an old version of `mypy` that was not compatible with all supported numpy versions. I bumped the mypy version in the CI to `1.20`, the latest release of `mypy v1`.

Two other mypy issues came up with the updated release:

1. Not all `ATS` objects have a `lineno` attributed, so I added a failsafe that returns `'unknown'` when there is no `lineno` attribute.
2. `numpy.typing.mypy_plugin` is now depreciated, so I removed it from `mypy.init`.

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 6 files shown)


IvanGrigorik (approved):
LG, thanks

PR #431, opened 2026-07-02
feat: handle multi-level scratch memory

### LINKED ISSUE #290: Initialize multiple levels of stratch memory.
Kokkos supports calling multiple levels of scratch memory as in the code below

```cpp
policy.set_scratch_size(0, PerTeam(bytes0))
      .set_scratch_size(1, PerTeam(bytes1));
```
but pykokkos throws an error when I call
```python
pk.TeamPolicy(league_size, team_size).set_scratch_size(0, pk.PerTeam(bytes0))
                                     .set_scratch_size(1, pk.PerTeam(bytes1))
```

Error:
```
warmup run
Traceback (most recent call last):
  File "/work/09661/gkk345/ls6/3dcapsules/python/development/gpu_timing.py", line 312, in <module>
    run()
  File "/work/09661/gkk345/ls6/3dcapsules/python/development/gpu_timing.py", line 106, in run
    p2g(gpu_pre)
  File "/work/09661/gkk345/ls6/3dcapsules/python/development/particle2grid/_p2g.py", line 28, in p2g
    pk.execute(
  File "/work/09661/gkk345/ls6/pykokkos/pykokkos/interface/parallel_dispatch.py", line 253, in execute
    runtime_singleton.runtime.run_workload(space, workload)
  File "/work/09661/gkk345/ls6/pykokkos/pykokkos/core/runtime.py", line 58, in run_workload
    members: PyKokkosMembers = self.compiler.compile_object(module_setup, space, km.is_uvm_enabled(), None, None, None, set())
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/09661/gkk345/ls6/pykokkos/pykokkos/core/compiler.py", line 185, in compile_object
    members = self.extract_members(entity, classtypes)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/09661/gkk345/ls6/pykokkos/pykokkos/core/compiler.py", line 364, in extract_members
    members.extract(entity, classtypes)
  File "/work/09661/gkk345/ls6/pykokkos/pykokkos/core/translators/members.py", line 100, in extract
    self.reduction_result_queue, self.timer_result_queue = self.get_queues(source, name, pk_import)
                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/09661/gkk345/ls6/pykokkos/pykokkos/core/translators/members.py", line 158, in get_queues
    node_visitor.visit(main)
  File "/work/09661/gkk345/ls6/miniconda3/envs/pyse/lib/python3.11/ast.py", line 418, in visit
    return visitor(node)
           ^^^^^^^^^^^^^
  File "/work/09661/gkk345/ls6/pykokkos/pykokkos/core/visitors/kokkosmain_visitor.py", line 38, in visit_FunctionDef
    run_body += serializer.serialize(self.visit(statement))
                                     ^^^^^^^^^^^^^^^^^^^^^
  File "/work/09661/gkk345/ls6/miniconda3/envs/pyse/lib/python3.11/ast.py", line 418, in visit
    return visitor(node)
           ^^^^^^^^^^^^^
  File "/work/09661/gkk345/ls6/pykokkos/pykokkos/core/visitors/pykokkos_visitor.py", line 510, in visit_Expr
    call: cppast.CallExpr = self.visit(node.value)
                            ^^^^^^^^^^^^^^^^^^^^^^
  File "/work/09661/gkk345/ls6/miniconda3/envs/pyse/lib/python3.11/ast.py", line 418, in visit
    return visitor(node)
           ^^^^^^^^^^^^^
  File "/work/09661/gkk345/ls6/pykokkos/pykokkos/core/visitors/kokkosmain_visitor.py", line 298, in visit_Call
    policy = self.add_space_to_policy(policy)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/09661/gkk345/ls6/pykokkos/pykokkos/core/visitors/kokkosmain_visitor.py", line 424, in add_space_to_policy
    policy_constructor.add_template_param(space)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'MemberCallExpr' object has no attribute 'add_template_param'
```

Machine: GPU-a100, cuda 11.4

WeiTaShao:
## Multi-level scratch memory

`multi_scratch_team.py` demonstrates how to request scratch memory on different Kokkos scratch levels from a PyKokkos workunit.

Use the optional third value in each `scratch` tuple to select the scratch level:

```python
@pk.workunit(
    scratch=[
        (int, lambda p: p.n_ints, 0),
        (float, lambda p: p.N, 1),
    ]
)
```

Each tuple has the form:

```python
(dtype, size_func, level)
```

Entries that use the same level are summed together. Entries that use different levels are passed to Kokkos as separate `set_scratch_size` calls.
Inside the workunit, use the matching `team_scratch(level)` call:

```python
scratch_idx = pk.ScratchView1D[int](team_member.team_scratch(0), n_ints)
scratch_val = pk.ScratchView1D[float](team_member.team_scratch(1), N)
```

The older two-value form still works and defaults to scratch level 0:

```python
scratch=[(float, lambda p: p.N)]
```


pykokkos/core/runtime.py
```diff
@@ -114,32 +149,42 @@ def apply_scratch_spec(workunit: Callable, policy: TeamPolicy, **kwargs) -> None
...
+                max_scratch = TeamPolicy.scratch_size_max(level)
                 if max_scratch > 0 and total_scratch_size > max_scratch:
                     raise ValueError(
                         f"Requested scratch size ({total_scratch_size} bytes) "
-                        f"exceeds maximum for level 0 ({max_scratch} bytes). "
+                        f"exceeds maximum for level {level} ({max_scratch} bytes). "
                         "Reduce scratch allocation or use a different level."
                     )
-            except (ImportError, AttributeError):
+            except (ImportError, AttributeError, RuntimeError):
```
kennykos: What causes Import, Attribute, or Runtime Errors here?

pykokkos/core/runtime.py
```diff
@@ -711,35 +756,34 @@ def get_policy_arguments(self, policy: ExecutionPolicy) -> Dict[str, Any]:
...
-                    # Direct size value (workunit case without wrapper)
-                    args["pk_scratch_size_is_per_team"] = (
-                        True  # Default to PerTeam for simple int
-                    )
-                    args["pk_scratch_size_value"] = int(policy.scratch_size_value)
+            scratch_sizes = dict(policy.scratch_sizes)
+            if policy.scratch_size_level is not None and not scratch_sizes:
+                scratch_sizes[policy.scratch_size_level] = policy.scratch_size_value
+
+            from pykokkos.interface.hierarchical import PerTeam, PerThread
```
kennykos: Can we import this at the top of the script? Importing at module initialization is much cheaper than importing at function call.

pykokkos/core/runtime.py
```diff
@@ -114,32 +149,42 @@ def apply_scratch_spec(workunit: Callable, policy: TeamPolicy, **kwargs) -> None
...
+            if total_scratch_size <= 0:
+                continue
 
-        if total_scratch_size > 0:
-            policy.scratch_size_level = 0
-            policy.scratch_size_value = PerTeam(total_scratch_size)
+            policy.scratch_sizes[level] = PerTeam(total_scratch_size)
             try:
-                max_scratch = TeamPolicy.scratch_size_max(0)
+                max_scratch = TeamPolicy.scratch_size_max(level)
```
kennykos: Is there a reason we use the generic `TeamPolicy` here and not the `policy` passed like so?
```suggestion
                max_scratch = policy.scratch_size_max(level)
```

kennykos (commented):
Great work! Does this require a rebuild of `PyKokkos base`? If so, is there any way we can rig it so we don't have to rebuild until this feature is needed?

kennykos:
Also, this will close #290

pykokkos/core/runtime.py
IvanGrigorik: @WeiTaShao, please always do global includes if that does not cause any circular import warnings/errors

pykokkos/core/keywords.py
```diff
@@ -25,3 +25,9 @@ class Keywords(Enum):
     ScratchSizeLevel = "pk_scratch_size_level"
     ScratchSizeValue = "pk_scratch_size_value"
     ScratchSizeIsPerTeam = "pk_scratch_size_is_per_team"
+    ScratchSizeLevel0Value = "pk_scratch_size_level_0_value"
+    ScratchSizeLevel0IsPerTeam = "pk_scratch_size_level_0_is_per_team"
+    ScratchSizeLevel0Enabled = "pk_scratch_size_level_0_enabled"
+    ScratchSizeLevel1Value = "pk_scratch_size_level_1_value"
+    ScratchSizeLevel1IsPerTeam = "pk_scratch_size_level_1_is_per_team"
+    ScratchSizeLevel1Enabled = "pk_scratch_size_level_1_enabled"
```
IvanGrigorik, line 33: Eh...
I didn't like the moment when I had to add it like another keyword. We should not abuse it so heavily.
Is there any other way to do it?

pykokkos/core/runtime.py
```diff
@@ -78,11 +78,42 @@ def _calculate_aligned_scratch_size(
...
+    """
+    if not isinstance(entry, tuple) or len(entry) not in (2, 3):
+        raise TypeError(
+            "scratch entries must be tuples of "
+            "(dtype, size_func) or (dtype, size_func, level)"
+        )
+
+    if len(entry) == 2:
+        dtype, size_func = entry
+        level = 0
```
IvanGrigorik, line 97: Add it to the doc, please
```
By default, the scratch level is set to 0
```
or something like  that

pykokkos/interface/execution_policy.py
```diff
@@ -198,6 +198,7 @@ def __init__(self, *args):
         self.vector_length: int = vector_length
         self.scratch_size_level: Optional[int] = None
         self.scratch_size_value = None
+        self.scratch_sizes: Dict[int, Any] = {}
```
IvanGrigorik: Why `Any`?

pykokkos/interface/views.py
```diff
@@ -1197,9 +1197,12 @@ def _calculate_scratch_size(type_size: int, *dims: int, alignment: int = 8) -> i
...
     def __class_getitem__(cls, item):
-        generic_alias = super().__class_getitem__(item)
-        generic_alias.type_param = item
-        return generic_alias
+        type_name = getattr(item, "__name__", str(item))
+        return type(
+            f"{cls.__name__}[{type_name}]",
+            (cls,),
+            {"__module__": cls.__module__, "type_param": item},
+        )
```
IvanGrigorik, line 1205: Thanks for refining that

examples/kokkos/multi_scratch_team.py
```diff
@@ -12,8 +12,8 @@ def init_data(i: int, view):
 #   scratch_val - N      float64 elements (passed as kwarg)
 @pk.workunit(
     scratch=[
-        (int, lambda p: p.n_ints),
-        (float, lambda p: p.N),
+        (int, lambda p: p.n_ints, 0),
+        (float, lambda p: p.N, 1),
     ]
```
IvanGrigorik, line 17: We should be somewhat reverse-compatible in this syntax.
So lets just make another example with level specificationn and leave this one as-is, with this syntax:
```python
@pk.workunit(
    scratch=[
        (int, lambda p: p.n_ints),
        (float, lambda p: p.N),
    ]
```

IvanGrigorik (commented):
Overall LG.
I would like to make another pass a bit later, since I am a bit concerned about a couple of moments

pykokkos/core/runtime.py
WeiTaShao: RuntimeError could happen when `TeamPolicy.scratch_size_max()` currently raises this explicitly when `kokkos.scratch_size_max` is not available.

WeiTaShao:
> Great work! Does this require a rebuild of `PyKokkos base`? If so, is there any way we can rig it so we don't have to rebuild until this feature is needed?

It is not required rebuilding PyKokkos base. The multi-level scratch support is implemented in the Python runtime and in the generated Kokkos call, using existing Kokkos `TeamPolicy::set_scratch_size(level, ...)` APIs.

The only part that may depend on the installed base binding is the optional `scratch_size_max(level) validation`. If an older PyKokkos base does not export `kokkos.scratch_size_max`, we should skip that validation rather than requiring a rebuild. 

The feature itself can still work without that check.

pykokkos/interface/execution_policy.py
WeiTaShao: `Any` was used because the current runtime path accepts several scratch-size representations before converting them into generated-kernel arguments:
`PerTeam`, `PerThread`, `plain int / np.integer`, and potentially other variable-like values passed through from policy state. But the `np.integer` is a runtime type, it may not be suitable to put type annotation

For the new decorator-based scratch path, the values stored in `policy.scratch_sizes` are currently `PerTeam(total_scratch_size)`. If we want the annotation to reflect both current behavior and likely future/manual policy support, a better type would be something like

```
Dict[int, Union[PerTeam, PerThread, int]]
```

pykokkos/core/keywords.py
WeiTaShao: A possible way is to replace them with small helper functions that generate the parameter names from the scratch level:

```
def scratch_size_value_name(level: int) -> str:
    return f"pk_scratch_size_level_{level}_value"

def scratch_size_is_per_team_name(level: int) -> str:
    return f"pk_scratch_size_level_{level}_is_per_team"

def scratch_size_enabled_name(level: int) -> str:
    return f"pk_scratch_size_level_{level}_enabled"
```

Then both`runtime.py` and `bindings.py` can use the same naming helpers instead of hardcoding six separate keyword enum members.

Would this be a better approach?

pykokkos/core/runtime.py
WeiTaShao: Done

pykokkos/core/runtime.py
WeiTaShao: Done

pykokkos/core/runtime.py
WeiTaShao: Done

examples/kokkos/multi_scratch_team.py
WeiTaShao: Done

pykokkos/core/runtime.py
IvanGrigorik: We should not just `pass` in this case. We need to warn the user about it
It also should not be the case since `kokkos.scratch_size_max` should always be available

docs/workunits.rst
```diff
@@ -56,6 +56,35 @@ Each work unit function can accept an arbitrary number of arguments
...
+
+   @pk.workunit(
+       scratch=[
+           (int, lambda p: p.n_ints, 0),
+           (float, lambda p: p.N, 1),
+       ]
+   )
+
+The tuple format is ``(dtype, size_func)`` or ``(dtype, size_func, level)``.
+
```
kennykos, line 87: This is great! Thanks.

pykokkos/core/runtime.py
WeiTaShao: Add a warning, can be removed if this case never occurs

pykokkos/core/keywords.py
IvanGrigorik: Let's leave it as-is. With functions it looks weird, like we have some shadow keyword.

pykokkos/interface/execution_policy.py
IvanGrigorik: Union looks better. Some class wrapper would look even better

docs/workunits.rst
```diff
@@ -56,6 +56,35 @@ Each work unit function can accept an arbitrary number of arguments
 after those required in the signature.  At the time of a pattern
 execution, the arguments have to be given as keyword arguments.
 
+Scratch Memory
+--------------
+
+Work units can request team scratch memory with the ``scratch`` argument to
```
IvanGrigorik: `Workunits`
Idk how to "suggest change" as Gabriel did that...

pykokkos/core/runtime.py
```diff
@@ -114,31 +150,44 @@ def apply_scratch_spec(workunit: Callable, policy: TeamPolicy, **kwargs) -> None
...
                     )
-            except (ImportError, AttributeError):
-                # backend may not expose scratch_size_max
-                pass
+            except (ImportError, AttributeError, RuntimeError) as exc:
+                warnings.warn(
+                    f"Unable to query scratch_size_max for scratch level {level}: "
+                    f"{exc}. Skipping scratch size validation.",
+                    RuntimeWarning,
+                )
```
IvanGrigorik: Let's just raise error

IvanGrigorik (commented):
Overall LG. 
Final changes + I'll make another pass a bit later

pykokkos/core/runtime.py
WeiTaShao: Done

pykokkos/interface/execution_policy.py
WeiTaShao: Done

docs/workunits.rst
kennykos:
```suggestion
`Workunits` can request team scratch memory with the ``scratch`` argument to
```

WeiTaShao: Done

IvanGrigorik (approved):
I tested it locally with an NVIDIA GPU.
LG!

PR #435, opened 2026-08-01
Fix nested workunits

WeiTaShao:
## Implicit helper functions

Standalone PyKokkos workunits can call compatible module-level Python functions without requiring the `@pk.function` decorator. PyKokkos discovers helper functions reachable from the workunit, recursively includes their same-module dependencies in translation, and excludes unrelated host-side functions.

```python
import pykokkos as pk


def my_printer(i: int) -> None:
    pk.printf("Index: %d\n", i)


@pk.workunit
def hello(i: int) -> None:
    my_printer(i)


pk.parallel_for("hello_loop", 10, hello)

### DIFF
--- a/pykokkos/core/compiler.py
+++ b/pykokkos/core/compiler.py
@@ -161,7 +161,9 @@ class Compiler:
             # Avoid fusing the ASTs before checking if it was already compiled
             entity, classtypes = self.fuse_objects(metadata, fuse_ASTs=False, **kwargs)
 
-        hash: str = self.members_hash(entity.path, entity.name, types_signature)
+        hash: str = self.members_hash(
+            entity.path, entity.name, module_setup.ast_signature, types_signature
+        )
 
         types_inferred: bool = updated_types is not None
         decorator_inferred: bool = updated_decorator is not None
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pykokkos/core/compiler.py [hunk 1]
```diff
@@ -393,21 +395,26 @@ def read_defaults(self) -> Optional[CompilationDefaults]:
         return defaults
 
     def members_hash(
-        self, path: List[str], name: str, types_signature: Optional[str]
+        self,
+        path: List[str],
+        name: str,
+        ast_signature: str,
```
IvanGrigorik, line 421: So...
Now you will recompile if ANYTHING in AST changed?
That's questionable.

tests/test_implicit_functions.py
IvanGrigorik, line 1: Please add a test of defining a function in one module (file) and using it from another module workunit.

IvanGrigorik (commented):
`self.add_called_functions` is highly questionable. 
I'm not sure if this is the easiest way to check which function is called.

Instead of gathering all possible functions (which is insanely heavy) we can check which function is called based on the current context directly from the AST translator.

---

And yes, don't remote `@pk.function` decorator nor ignore it. It will be a pain if we will try to visit each function without looking at its source of it.

